PR #3, opened 2024-02-27
ENH: add dataclasses for data model

tangkong:
## Description
* Adds basic datacalsses for Entry, Parameter, Value, Collection, and Snapshot
* Adds initial validation tests, and a sample database fixture

## Motivation and Context
Continue fleshing out structure of this package's backend

### Notes on validation
I've chosen `apischema` here since it is the beast we know, and has similar functionality to pydantic.  `apischema` performs type validation happens on serialization/deserialization, but not outside of that.  If we are to be constructing these dataclasses we'll want to validate them before attempting to save them to a backend (whether serialized or not).  

To help with that, the `.validate()` method has been added.  This calls the `apischema` validation routine.  Adding routines  can be done with the `@validator` decorator, quite easily.

### In pursuit of lazy loading
This section is also the "Why hint `Union[UUID, Entry]` everwhere?" section

I expect lazy loading will become a requirement, especially as database size grows.  I imagine a world where we have a huge root tree structure, with deep nesting of collections and snapshots.  In this case, we wouldn't want to pull the whole database's contents with every app load.  

Instead we might first grab the metadata information (name, description, creation time), then fill its children when their information is requested.  Exactly how we accomplish this should be fleshed out in the client/backend implementations, but the data-model would have to allow "unfilled" references to `Entry` objects.  

## How Has This Been Tested?
Added a few tests, need to flesh them out a bit

This PR, soon pre-release notes

## Pre-merge checklist


tangkong:
The guts of this might be ok, but after some deeper thinking about this, I'm not sure this is the data schema we want to go with.  Moving to draft for now while conversations happen elsewhere

PR #24, opened 2024-05-01
ENH: initial skeleton for Client and Backend

### LINKED ISSUE #13: ENH: Build Core Backend interface
Establish the base class for `Backend`s.  

Likely has methods limited to the basic CRUD operations.

### LINKED ISSUE #23: Create Client Skeleton

tangkong:
## Description
Skeleton structure for Client and Backend. 
To-Do:
- figure out if we can add any reasonable tests at this point
- think about whether a "root" node is reasonable / necessary
  - My pitch for this is that we need a way to get the top-level `Entry`'s, to populate an initial preview of the database (tree view).  As it stands there's no way to distinguish a "parent" from a "child".
  - A simple way of doing this might be to just have a simple `Root` container that holds the top-level `Entry`'s.  This could have a static UUID so we can always grab it.
- add exceptions module?

## Motivation and Context
We want to broadly agree on the available methods so we can do other work.

closes #23
closes #13 

## How Has This Been Tested?
Adding some tests, we need to add more

This PR, design doc

## Pre-merge checklist

### DIFF
--- a/superscore/backends/core.py
+++ b/superscore/backends/core.py
@@ -1,10 +1,10 @@
 """
 Base superscore data storage backend interface
 """
-from typing import Any, Generator
+from typing import Generator
 from uuid import UUID
 
-from ..model import Entry
+from superscore.model import Entry
 
 
 class _Backend:
@@ -16,7 +16,10 @@ class _Backend:
         raise NotImplementedError
 
     def save_entry(self, entry: Entry):
-        """Save ``entry`` into the database"""
+        """
+        Save ``entry`` into the database
+        Throws BackendError if ``entry`` already exists
+        """
         raise NotImplementedError
 
     def delete_entry(self, entry: Entry) -> None:
@@ -32,6 +35,6 @@ class _Backend:
         """
         raise NotImplementedError
 
-    def search(self, **search_kwargs) -> Generator[Any, None, None]:
+    def search(self, **search_kwargs) -> Generator[Entry, None, None]:
         """Yield a Entry objects corresponding matching ``search_kwargs``"""
         raise NotImplementedError
--- a/superscore/backends/core.py
+++ b/superscore/backends/core.py
@@ -1,0 +1,37 @@
+"""
+Base superscore data storage backend interface
+"""
+from typing import Any, Generator
+from uuid import UUID
+
+from ..model import Entry
+
+
+class _Backend:
+    """
+    Base class for data storage backend.
+    """
+    def get_entry(self, meta_id: UUID) -> Entry:
+        """Get entry with ``meta_id``."""
+        raise NotImplementedError
+
+    def save_entry(self, entry: Entry):
+        """Save ``entry`` into the database"""
+        raise NotImplementedError
+
+    def delete_entry(self, entry: Entry) -> None:
+        """
+        Delete ``entry`` from the system (all instances)
+        """
+        raise NotImplementedError
+
+    def update_entry(self, entry: Entry) -> None:
+        """
+        Update ``entry`` in the backend.
+        Throws BackendError if ``entry`` does not already exist
+        """
+        raise NotImplementedError
+
+    def search(self, **search_kwargs) -> Generator[Any, None, None]:
+        """Yield a Entry objects corresponding matching ``search_kwargs``"""
+        raise NotImplementedError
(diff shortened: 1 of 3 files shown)


tangkong:
Honestly not much here, but it's a start.  Better to keep additions small I guess

superscore/backends/core.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def update_entry(self, entry: Entry) -> None:
+        """
+        Update ``entry`` in the backend.
+        Throws BackendError if ``entry`` does not already exist
+        """
+        raise NotImplementedError
+
+    def search(self, **search_kwargs) -> Generator[Any, None, None]:
+        """Yield a Entry objects corresponding matching ``search_kwargs``"""
```
ZLLentz, line 36: I'm assuming that each search needs to be backend-specific, but are there any backend-agnostic API decisions we'd want to make here about how the search_kwargs or the output should behave?

For example, should search_kwargs be AND-ed or OR-ed together? Should the generator always yield Entry objects? Are there any assumptions we can make about regex support for search terms? etc.

(I don't immediately have answers as I go through this review coming up on 5pm)

superscore/backends/core.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    def get_entry(self, meta_id: UUID) -> Entry:
+        """Get entry with ``meta_id``."""
+        raise NotImplementedError
+
+    def save_entry(self, entry: Entry):
+        """Save ``entry`` into the database"""
+        raise NotImplementedError
+
+    def delete_entry(self, entry: Entry) -> None:
```
ZLLentz, line 22: Would this require the entire `Entry` object, or would the `meta_id` suffice?

superscore/backends/core.py
```diff
@@ -0,0 +1,37 @@
...
+class _Backend:
+    """
+    Base class for data storage backend.
+    """
+    def get_entry(self, meta_id: UUID) -> Entry:
+        """Get entry with ``meta_id``."""
+        raise NotImplementedError
+
+    def save_entry(self, entry: Entry):
+        """Save ``entry`` into the database"""
```
ZLLentz, line 19: Would this be the opposite of `update_entry`, in that it would error if the entry's meta_id already existed?

superscore/client.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, backend=None, **kwargs) -> None:
+        # if backend is None, startup default filestore backend
+        return
+
+    @classmethod
+    def from_config(cls, cfg=None):
+        raise NotImplementedError
+
+    def search(self, **post) -> List[Entry]:
```
ZLLentz, line 19: This signature here is at odds with the implementation, which returns a generator.

ZLLentz (commented):
Some just-before-5pm thoughts

superscore/backends/core.py
tangkong: I think so.  I think that distinction makes sense for the use cases we've been talking about (allow most users to save new entries, but only allow some users to modify existing entries)

I should make that more clear in the docstring like I did for `update_entry`

superscore/backends/core.py
tangkong: `meta_id` would suffice, maybe I wanted to keep the signatures similar for ease of memory.

superscore/backends/core.py
```diff
@@ -0,0 +1,37 @@
...
+        raise NotImplementedError
+
+    def update_entry(self, entry: Entry) -> None:
+        """
+        Update ``entry`` in the backend.
+        Throws BackendError if ``entry`` does not already exist
+        """
+        raise NotImplementedError
+
+    def search(self, **search_kwargs) -> Generator[Any, None, None]:
```
shilorigins, line 35: Do we expect the generator to yield anything other than `Entry`s?

superscore/client.py
tangkong: I'll keep this as a generator for now to remind us to be lazy when we can, but when this is implemented I'm sure we'll have more discussion

superscore/backends/core.py
tangkong: likely not 👍

shilorigins (approved):

ZLLentz (approved):

PR #25, opened 2024-05-10
ENH: Add Filestore backend

tangkong:
## Description
Still some work to be done, but I think I should open this up to review before I go too off the rails.

- Adds Filestore Backend
- Adds Root to the model definitions for holding top-level entries
- Added `UUID` to potentially lazy fields.

## Motivation and Context
Much of this is inspired from lessons learned from happi, forgive any vestigial bits that have made their way in.  
### What's going on here?
Internally, the FilestoreBackend holds one data structure in addition to the `Root`:
* an entry cache, mapping uuid to lazy-ified Entry

The entry cache is used to quickly grab an `Entry` or search without recursing through the tree (avoiding double hits).  With each CRUD operation, the backend uses this cache to reconstruct the `Root` object, filling in uuid references where they exist. 

In the future, I expect us to keep a sort of uuid-link-map, mapping a uuid to other uuids that reference it.  
When other backends delete an entry, we will need this information to make sure we dereference the other entries properly.  However since the FilestoreBackend reconstructs the entire root with every modification, it's not strictly needed

### Why did you add a `Root` object?
In any backend context, we need some way to know what `Entry`s are at the top level of the tree structure.  While in the Filestore backend case we might store the `Entry`s in a "filled" state (with full objects as children recursively), if we ever want to flatten this structure we will need to keep track of the top-level items.  

A simple "Root" object (holding only a list of `Entry`s) seemed like a simple way to do this.

### Why did `@as_tagged_union` come back?
It may not be totally needed (we could specify each of the `Entry` subclasses in `Root`), but I think it helps the readability of the serialized output.  Also there are issues with Value -> Setpoint/Readback, the nested tagged-union has been causing issues with serialization round trips

### UUIDs are back
Admittedly we could probably run through the entire filestore database every time we searched for an Entry, but I used this as a potential example of how we might implement entry caching, lazy loading, and tree reconstruction.  The rough idea for the filestore backend side is to regenerate and save the json db with every action.  

## How Has This Been Tested?
Added some unit tests, more to come.

This PR

## Pre-merge checklist


tangkong:
I'll attempt to describe my current sticking point so I can come back to it when I have more time.  Our schema has everything inheriting from `Entry`.  A subclass `Value` is then subclassed again into `Setpoint` and `Readback`, with the current intent to use all three.  Currently this creates issues where apischema cannot distinguish what type is desired.

* If we hint `Value`, but supply a `Setpoint` or `Readback`, we will lose all subclass information since we cast the object to a `Value`
* If we hint `Setpoint` or `Readback`, we can no longer use `Value`, since it is not technically a `Setpoint` or `Readback` (a square is a rectangle but a rectangle is not a square)
* If we hint all three, apischema short-circuits at the first applicable hint. 

I think this behavior all makes sense, given the ambiguity of these classes.  I see two main options:
* Stop using `Value`, and leave it act only as an abstract base class of sorts.  Then hint only `Setpoint`/`Readback`
* Make `Setpoint`/`Readback` not inherit from `Value`, and duplicate the fields.  Then hint all 3.

I tried looking around for apischema-voodoo that would let us use a tagged_union of a tagged_union, but didn't anything super simple.

tangkong:
This has gone on long enough.  I'll request reviews and give it another once-over tomorrow with fresher eyes.

superscore/utils.py
```diff
@@ -6,3 +7,22 @@
...
+    basedir : str
+        If path is not absolute already, build an abspath
+        with path starting here.
+    path : str
+        The path to convert to absolute.
+    """
+    path = os.path.expanduser(path)
+    if not os.path.isabs(path):
+        return os.path.abspath(os.path.join(basedir, path))
+    return path
```
ZLLentz, line 28: This utils function is a bit weird to me. In a lot of cases it's not necessary to provide a basedir to figure out a full absolute path, and in many cases this input arg is ignored entirely. Maybe it has specific utility in suggesting directories to save in?

superscore/backends/filestore.py
```diff
@@ -0,0 +1,305 @@
...
+        """Delete meta_id from the system (all instances)"""
+        with _load_and_store_context(self) as db:
+            db.pop(entry.uuid, None)
+
+    def search(self, **search_kwargs) -> Generator[Entry, None, None]:
+        """
+        Search for an entry that matches ``search_kwargs``.
+        Currently does not support partial matches.
+        """
+        for entry in self._entry_cache.values():
```
ZLLentz, line 288: self._entry_cache is not guaranteed to be up-to-date here, since it is only rebuilt prior to modification operations using the context manager, not after them.

superscore/errors.py
ZLLentz, line 1: This needs to be merged with the errors from the other PR

superscore/tests/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+
+import pytest
+
+from superscore.backends.core import _Backend
+from superscore.backends.filestore import FilestoreBackend
+from superscore.model import Collection, Parameter, Root, Setpoint, Snapshot
+
+
+@pytest.fixture(scope='function')
+def sample_database() -> Root:
```
ZLLentz, line 13: Is this at odds with the fixture defined in https://github.com/pcdshub/superscore/blob/beb6977ba57dd42e455beadc60e8079bac65f1c6/superscore/tests/conftest.py#L8 or do they serve different purposes?

ZLLentz (commented):
Partial review, I'm out on Friday but I can continue next week if needed

superscore/tests/conftest.py
tangkong: I think they can coexist.  Having multiple test backends can't hurt right?

superscore/backends/filestore.py
tangkong: Good catch, I'll wrap this as well

tangkong:
An example serialized file if we remove `@as_tagged_union` and just hint all the types of `Entry`: losing the class name 

<details><summary>Details</summary>
<p>

```json
{
  "meta_id": "a28cd77d-cc92-46cc-90cb-758f0f36f041",
  "entries": [
    {
      "uuid": "dfce8fb3-0a13-4fb3-b34e-aa7752decf26",
      "description": "parameter 1 in root",
      "creation_time": "2024-05-16T23:52:28.879185+00:00",
      "pv_name": "MY:MOTOR:mtr1.ACCL",
      "abs_tolerance": null,
      "rel_tolerance": null,
      "readback": null,
      "read_only": false
    },
    {
      "uuid": "0c566d5a-5ec8-4e7c-9d65-625a0d44185e",
      "description": "",
      "creation_time": "2024-05-16T23:52:28.879216+00:00",
      "pv_name": "MY:MOTOR:mtr1.ACCL",
      "data": 2,
      "status": 18,
      "severity": 4,
      "readback": null
    },
    {
      "uuid": "24fc4d5d-545f-4b54-9a7d-233866f0cb0d",
      "description": "collection 1 defining some motor fields",
      "creation_time": "2024-05-16T23:52:28.879236+00:00",
      "title": "collection 1",
      "children": [
        {
          "uuid": "067e4eb3-b745-49b9-97ba-669c17b7a1a3",
          "description": "motor field ACCL",
          "creation_time": "2024-05-16T23:52:28.879277+00:00",
          "pv_name": "MY:PREFIX:mtr1.ACCL",
          "abs_tolerance": null,
          "rel_tolerance": null,
          "readback": null,
          "read_only": false
        },
        {
          "uuid": "c1ea420c-735f-4299-811c-4726863df69c",
          "description": "motor field VELO",
          "creation_time": "2024-05-16T23:52:28.879314+00:00",
          "pv_name": "MY:PREFIX:mtr1.VELO",
          "abs_tolerance": null,
          "rel_tolerance": null,
          "readback": null,
          "read_only": false
        },
        {
          "uuid": "68581b2c-a69e-4695-94ef-2e6de2c81605",
          "description": "motor field PREC",
          "creation_time": "2024-05-16T23:52:28.879350+00:00",
          "pv_name": "MY:PREFIX:mtr1.PREC",
          "abs_tolerance": null,
          "rel_tolerance": null,
          "readback": null,
          "read_only": false
        }
      ],
      "tags": []
    },
    {
      "uuid": "32e5d77c-350e-4842-807a-5e281f1a8023",
      "description": "Snapshot 1 created from collection 1",
      "creation_time": "2024-05-16T23:52:28.879255+00:00",
      "title": "snapshot 1",
      "origin_collection": null,
      "children": [
        {
          "uuid": "f71c5363-a7e0-479e-b748-f034eb9f75f4",
          "description": "",
          "creation_time": "2024-05-16T23:52:28.879295+00:00",
          "pv_name": "MY:PREFIX:mtr1.ACCL",
          "data": 2,
          "status": 18,
          "severity": 4,
          "readback": null
        },
        {
          "uuid": "74405890-5fbb-4c44-916a-1ffc728d8e70",
          "description": "",
          "creation_time": "2024-05-16T23:52:28.879331+00:00",
          "pv_name": "MY:PREFIX:mtr1.VELO",
          "data": 2,
          "status": 18,
          "severity": 4,
          "readback": null
        },
        {
          "uuid": "0fec8587-6cc7-41e8-823f-11b43f3ed3d9",
          "description": "",
          "creation_time": "2024-05-16T23:52:28.879367+00:00",
          "pv_name": "MY:PREFIX:mtr1.PREC",
          "data": 6,
          "status": 18,
          "severity": 4,
          "readback": null
        }
      ],
      "tags": [],
      "meta_pvs": []
    }
  ]
}
```

</p>
</details>

superscore/model.py
```diff
@@ -50,20 +53,31 @@ class Tag(Flag):
...
     creation_time: datetime = field(default_factory=utcnow)
 
+    def swap_to_uuids(self) -> List[Union[Entry, UUID]]:
+        """
+        Swap to UUID references where relevant.
+
+        Returns a list of objects referenced by this Entry,
+        either an Entry's that has been swapped to UUIDs,
+        or a UUID if it was already filled with a UUID.
+        """
```
ZLLentz, line 73: In both places you've implemented this, this method also mutates the `Entry` via `self.children = new_children`. I think we either shouldn't do that, or we should modify the docstring here to require this behavior so that it is clear and consistent.

superscore/model.py
```diff
@@ -73,20 +87,41 @@ class Parameter(Entry):
...
-    pv_name: str = ''
+    pv_name: str = ""
     data: Optional[AnyEpicsType] = None
     status: Status = Status.UDF
     severity: Severity = Severity.INVALID
 
 
 @dataclass
-class Setpoint(Value):
+class Setpoint(Entry):
```
ZLLentz, line 97: What is `Value` used for after this PR?

superscore/tests/conftest.py
ZLLentz: If they both exist as in-memory databases for testing and if they disagree on how to set up such a database, it's possible that we can get confused later.

superscore/model.py
```diff
@@ -8,10 +9,12 @@
 from typing import ClassVar, List, Optional, Set, Union
 from uuid import UUID, uuid4
 
+from superscore.serialization import as_tagged_union
 from superscore.type_hints import AnyEpicsType
 from superscore.utils import utcnow
 
 logger = logging.getLogger(__name__)
+_root_uuid = _root_uuid = UUID("a28cd77d-cc92-46cc-90cb-758f0f36f041")
```
ZLLentz, line 17: What's the idea behind this root uuid? It looks like it is static and is placed into every file exported by the filestore backend?

ZLLentz (commented):
Some more thoughts from EOD reviewing

superscore/tests/conftest.py
tangkong: This is fair.  I would can get behind unifying how we set them up, rather than avoiding having multiple backends

superscore/model.py
tangkong: It is static.  The concept behind `Root` is to have a single entry in the database with this UUID.  This entry holds the top level `Entry`s, for constructing a tree-view for example.  

Otherwise there would be no way to determine which `Entry`s are at the root level, and distinguish them from child `Entry`s

superscore/model.py
tangkong: I can certainly update the docstring to be more explicit, but I think it's worth discussing whether you all agree with the way I implemented this.  This is somewhat central to my pass at lazy loading.  

In the world where we don't mutate `Entry`, we would return a copy of the uuid-ified `Entry`?  If people prefer that I'm happy to go that route.

superscore/model.py
tangkong: To me there are `Value`s that are not `Setpoint`-`Readback` pairs and I wanted to leave that option open.  If people are ok with having `readback=None` for that type of `Value`, we can remove `Value`

superscore/backends/filestore.py
```diff
@@ -0,0 +1,306 @@
...
+                shutil.copymode(self.path, temp_path)
+            shutil.move(temp_path, self.path)
+        except BaseException as ex:
+            logger.debug('JSON db move failed: %s', ex, exc_info=ex)
+            # remove temporary file
+            if os.path.exists(temp_path):
+                os.remove(temp_path)
+            raise
+
+    def _temp_path(self) -> str:
```
shilorigins, line 225: I'd recommend looking at `tempfile` in the standard lib, which provides similar functionality in a context manager.

superscore/backends/filestore.py
```diff
@@ -0,0 +1,306 @@
...
+        with _load_and_store_context(self) as db:
+            for entry in db.values():
+                match = (getattr(entry, key, None) == value
+                         for key, value in search_kwargs.items())
+                if all(match):
+                    yield entry
+
+
+@contextlib.contextmanager
+def _load_and_store_context(
```
shilorigins, line 297: I think it'd be cleaner to define a context manager using `FilestoreBackend.__enter__` and `.__exit__`.

If we want to support multiple contexts, then I think this still belongs inside `FilestoreBackend`.

superscore/backends/filestore.py
```diff
@@ -0,0 +1,306 @@
...
+
+    @property
+    def root(self) -> Root:
+        """Refresh the cache and return the root object"""
+        with _load_and_store_context(self):
+            return self._root
+
+    def get_entry(self, uuid: UUID) -> Entry:
+        """Return the entry with ``uuid``"""
+        with _load_and_store_context(self) as db:
```
shilorigins, line 255: With these CRUD methods using a context manager that ends with `.store()`, we're going to write to the file system after every individual call. Our CRUD methods are even defined to accept single entries, so this will introduce a lot of overhead when making multiple changes.

As an alternative, should we remove the context manager from these methods and rely on the client to organize its calls into transactions?  Or we could add support for groups of entries, either in these method signatures or separate bulk methods.

superscore/backends/filestore.py
shilorigins, line 297: Another context to consider is one where we don't call `backend.store()`. This could be useful to avoid writing to the filesystem when no changes were made, such as in `_Backend.get()`.

superscore/tests/conftest.py
shilorigins: I support having multiple as long as the interface is consistent.

superscore/model.py
shilorigins: I prefer having `Setpoint` and `Readback` inherit from `Value`, but I know that that's causing issues with your serialization. If we can't resolve that issue, then I'm fine with removing `Value`.

superscore/model.py
shilorigins: Is this a convenient PR to introduce lazy loading, or could we postpone it to a separate PR? I'd like to take a crack at my idea as well, but I need a functional out-of-memory backend. I know your implementation is already in here, so no worries if it'd be tedious to extract.

superscore/tests/conftest.py
```diff
@@ -295,3 +301,66 @@ def linac_backend():
     )
 
     return TestBackend([root_col])
+
+
+@pytest.fixture(scope='function')
```
shilorigins, line 306: `function` is the default fixture scope, so can remove the arg from these. Alternatively, we can add the arg to all fixtures to make scope explicit.

superscore/model.py
ZLLentz: Ok, it stuck out as me as weird since I assumed the root uuid was a database unique identifier or something since it gets included in the serialized data file. I guess it's not a problem if every filestore database has an extra a28cd in it.

superscore/backends/filestore.py
```diff
@@ -0,0 +1,306 @@
...
+    """
+    Filestore configuration backend.
+
+    Holds an entry cache, filled when a file is loaded.
+    All CRUD operations reconstruct the Root object and save a new version of
+    the database.
+    File storage is a json file containing serialized model dataclasses.
+    """
+    _entry_cache: Dict[UUID, Entry] = {}
+    _root: Root
```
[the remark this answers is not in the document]
tangkong: So `_root` here is just a type hint, I'm not actually assigning it anything at this point.  This won't be shared across different instances of `FilestoreBackend`.

entry_cache on the other hand is defined like a class variable would be, I should change that

superscore/model.py
tangkong: So technically I haven't fully implemented lazy loading here.  In fact this backend is probably too "hard working" (what's the opposite of lazy in this context?).  

You should totally feel free to give your idea a spin.  I think lazy loading will look different for different backends

superscore/backends/filestore.py
tangkong: The benefit of rolling our own here is that in the case of an interruption, the temp file will continue to exist and can be recovered.

superscore/backends/filestore.py
tangkong: These are entirely valid optimization issues that I intentionally omitted from this backend.  I think it's ok to let this backend take a performance hit, particularly for transparency during testing.  If we want to optimize this backend later I think that's fine too, I just don't see it as particularly high priority for an MVP.
 
I think the backends should handle transaction batching, and expect that handling to be backend specific.  That being said I don't have a clear vision of how this would work for a mongo backend for example.

superscore/backends/filestore.py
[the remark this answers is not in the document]
shilorigins: Got it. I didn't realize you could typehint an attribute without actually declaring the attribute.

superscore/model.py
```diff
@@ -97,10 +132,34 @@ class Readback(Value):
...
 
+    @classmethod
+    def from_parameter(
+        cls,
+        origin: Parameter,
+        data: AnyEpicsType,
+        status: Status = Status.UDF,
+        severity: Severity = Severity.INVALID,
+        timeout: Optional[float] = None,
+    ) -> Setpoint:
```
shilorigins, line 151:
```suggestion
    ) -> Readback
```

superscore/backends/filestore.py
tangkong: I played around with `tempfile.NamedTemporaryFile`, which seemed promising at first.  I wasn't able to easily incorporate it into the backend, and there is some OS-dependent behavior I'd like to avoid.  

This file auto-deletes itself after exiting the context manager, which seems convenient.  Except for the fact that we attempt to move the temp file to the existing file location.  After doing this the tmp file no longer exists and the context manager throws.  Copy transactions are not guaranteed to be atomic, so I actually think we might have to roll our own here.

superscore/backends/filestore.py
tangkong: An attempt at summarizing the discussion from Tuesday's dev-discussion, based on my failing memory:
- I'm all for moving this into the FilestoreBackend
- The idea behind having `_Backend.__enter__` and `_Backend.__exit__` would allow for the client to coordinate transaction batching, but is something that complicates the user-facing API of the `Client`.
- I argued for having the `_Backend` handle batching, since optimal transaction grouping is likely backend dependent and I didn't want users to worry about this on the `Client` side
- Devan pointed out that this would require the `_Backend` to hold and process some sort of transaction queue, making it "stateful" in a way that's potentially also confusing and not consistently defined.

I don't think we came to conclusion on all these points, and I think they warrant more discussion / thought.

superscore/model.py
tangkong: I'll remove Value then.  I think it's unfortunate that I can't get the tagged-unions-of-tagged-unions to work, but from a usability standpoint I don't think there's much of a difference.  It's not the most DRY code in the world but I'd argue there are more offensive things out there...

superscore/tests/conftest.py
tangkong: captured some thoughts in: https://github.com/pcdshub/superscore/issues/33

shilorigins (approved):

ZLLentz (approved):

PR #27, opened 2024-05-14
Add initial test infrastructure: testing backend and general-use test fixture

### LINKED ISSUE #30: Raise error in _Backend.delete_entry when arg Entry data doesn't match data in the backend
### Discussed in https://github.com/pcdshub/superscore/discussions/29

<div type='discussions-op-text'>

<sup>Originally posted by **shilorigins** May 15, 2024</sup>
`_Backend.delete_entry` requires an `Entry` as input, which raises a question: what if the entry's UUID exists in the backend, but the entry itself has different data?  Should this raise an error because the arg entry "doesn't exist"?  If not, we should really change the signature to accept a UUID to remove this ambiguity.</div>

shilorigins:
## Description

Add `TestBackend`: stores `Entry`s in-memory for easy modification and fast access.  Direct object manipulation avoids the need to implement `update_entry` or `add_entry` at this time.

Add `linac_backend` test fixture: provides a large-ish tree of `Collection`s and `Parameter`s for tests where a larger tree is needed, such as recursive functions.  Can be modified within a test to provide different test conditions. Roughly follows the AD LINAC structure for ease of navigation.

Edit `_Backend.delete_entry` docstring to specify that the method raises an error when the `Entry` passed in as an argument is out-of-sync with the backend.  Being out-of-sync means that an `Entry` with a matching UUID exists in the backend, but its data is different. Closes #30 

Introduce custom exceptions: specifically `BackendError`, `EntryExistsError`, and `EntryNotFoundError`.
## Motivation and Context

`TestBackend` provides a way for tests to fetch and modify particular `Entry`s when using fixtures.  This enables us to reuse general-use fixtures rather than creating new fixtures tailors to different tests.

Building and storing the tree in-memory decouples the fixture from (de)serialization, whose format may change over time.
## How Has This Been Tested?

I used this fixture for testing #28 , but I split it into its own PR to keep both more manageable.

I've also included tests for each of `TestBackend`'s four CRUD methods.

<details markdown="block"><summary>Test coverage</summary>

```
Name                    Stmts   Miss  Cover
-------------------------------------------
__init__.py                 2      0   100%
__main__.py                 2      2     0%
_version.py                11     11     0%
backends/__init__.py        0      0   100%
backends/core.py           14      5    64%
backends/test.py           36      0   100%
bin/__init__.py             2      2     0%
bin/help.py                11     11     0%
bin/main.py                51     51     0%
bin/ui.py                  13     13     0%
client.py                  22     22     0%
errors.py                   6      0   100%
model.py                   80      0   100%
tests/__init__.py           0      0   100%
tests/conftest.py          38      0   100%
tests/test_backend.py      34      0   100%
tests/test_model.py        32      0   100%
tests/test_window.py        5      0   100%
type_hints.py               2      0   100%
utils.py                    5      0   100%
version.py                 27     17    37%
widgets/__init__.py         0      0   100%
widgets/core.py             5      0   100%
widgets/window.py           5      0   100%
-------------------------------------------
TOTAL                     403    134    67%
```
</details>

<details><summary>Test durations</summary>

```
================================ slowest durations =================================
0.07s setup    superscore/tests/test_window.py::test_main_window
0.03s call     superscore/tests/test_window.py::test_main_window
0.01s call     superscore/tests/test_model.py::test_serialize_collection_roundtrip
0.01s call     superscore/tests/test_model.py::test_serialize_snapshot_roundtrip

(17 durations < 0.005s hidden.  Use -vv to show these durations.)
```
</details> 

This PR, plus some discussion in #29 

## Pre-merge checklist

### DIFF
--- a/superscore/backends/test.py
+++ b/superscore/backends/test.py
@@ -1,0 +1,48 @@
+"""
+Backend that manipulates Entries in-memory for testing purposes.
+"""
+from typing import List
+from uuid import UUID
+
+from superscore.backends.core import _Backend
+from superscore.errors import (BackendError, EntryExistsError,
+                               EntryNotFoundError)
+from superscore.model import Collection, Entry, Snapshot
+
+
+class TestBackend(_Backend):
+    """Backend that manipulates Entries in-memory, for testing purposes."""
+    def __init__(self, data: List[Entry]):
+        self.data = data
+
+    def save_entry(self, entry: Entry) -> None:
+        try:
+            self.get_entry(entry.uuid)
+            raise EntryExistsError(f"Entry {entry.uuid} already exists")
+        except EntryNotFoundError:
+            self.data.append(entry)
+
+    def get_entry(self, uuid: UUID) -> Entry:
+        stack = self.data.copy()
+        while len(stack) > 0:
+            entry = stack.pop()
+            if entry.uuid == uuid:
+                return entry
+            if isinstance(entry, Collection) or isinstance(entry, Snapshot):
+                stack.extend(entry.children)
+        raise EntryNotFoundError(f"Entry {entry.uuid} could not be found")
+
+    def update_entry(self, entry: Entry) -> None:
+        original = self.get_entry(entry.uuid)
+        original.__dict__ = entry.__dict__
+
+    def delete_entry(self, to_delete: Entry) -> None:
+        stack = [self.data.copy()]
+        while len(stack) > 0:
+            children = stack.pop()
+            for entry in children.copy():
+                if entry == to_delete:
+                    children.remove(entry)
+                elif entry.uuid == to_delete.uuid:
+                    raise BackendError(f"Can't delete: entry {to_delete.uuid} is out of sync with the version in the backend")
+            stack.extend([entry.children for entry in children if isinstance(entry, Collection) or isinstance(entry, Snapshot)])
--- a/docs/source/upcoming_release_notes/27-Add_testing_backend_and_linac_fixture.rst
+++ b/docs/source/upcoming_release_notes/27-Add_testing_backend_and_linac_fixture.rst
@@ -1,0 +1,25 @@
+27 Add testing backend and linac fixture
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- Add custom exceptions: BackendError, FileExistsError, FileNotFoundError
+- Add TestBackend: an in-memory backend for testing purposes
+- Add LINAC Collection test fixture
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- Document that _Backend.delete_entry raises error when entry is out-of-sync with backend
+- Document that _Backend.get_entry raises error when entry can't be found
+
+Contributors
+------------
+- shilorigins
--- a/superscore/backends/core.py
+++ b/superscore/backends/core.py
@@ -12,13 +12,16 @@ class _Backend:
     Base class for data storage backend.
     """
     def get_entry(self, meta_id: UUID) -> Entry:
-        """Get entry with ``meta_id``."""
+        """
+        Get entry with ``meta_id``
+        Throws EntryNotFoundError
+        """
         raise NotImplementedError
 
     def save_entry(self, entry: Entry):
         """
         Save ``entry`` into the database
-        Throws BackendError if ``entry`` already exists
+        Throws EntryExistsError
         """
         raise NotImplementedError
 
@@ -31,7 +34,7 @@ class _Backend:
     def update_entry(self, entry: Entry) -> None:
         """
         Update ``entry`` in the backend.
-        Throws BackendError if ``entry`` does not already exist
+        Throws EntryNotFoundError
         """
         raise NotImplementedError
 
--- a/superscore/backends/core.py
+++ b/superscore/backends/core.py
@@ -28,6 +28,8 @@ class _Backend:
     def delete_entry(self, entry: Entry) -> None:
         """
         Delete ``entry`` from the system (all instances)
+        Throws BackendError if backend contains an entry with the same uuid as ``entry``
+        but different data
         """
         raise NotImplementedError
 
--- a/superscore/errors.py
+++ b/superscore/errors.py
@@ -1,0 +1,16 @@
+"""Custom Exceptions"""
+
+
+class BackendError(Exception):
+    """Raised when a Backend data operation fails"""
+    pass
+
+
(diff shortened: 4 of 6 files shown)


shilorigins:
Would it be fine to override `E251 unexpected spaces around keyword / parameter equals` for `tests/conftest.py`?  Since the constructor signatures are long, I like splitting them into multiple lines, where spaces around the `=` look better.

superscore/backends/test.py
```diff
@@ -0,0 +1,24 @@
...
+Backend that manipulates Entries in-memory for testing purposes.
+"""
+from typing import List
+from uuid import UUID
+
+from superscore.backends.core import _Backend
+from superscore.model import Collection, Entry, Snapshot
+
+
+class TestBackend(_Backend):
```
tangkong, line 11: You didn't want to implement the other `_Backend` CRUD methods?

tangkong (commented):
I have some more test infrastructure in my FilestoreBackend PR, that this should work nicely with hopefully

superscore/backends/test.py
shilorigins: I thought we'd add them as needed, but there's no harm in doing them now.

ZLLentz:
I assumed the intent of E251 was to differentiate between variable assignments and function call args, which is probably even more important than usual if things are spread out across many lines.

superscore/backends/test.py
```diff
@@ -0,0 +1,24 @@
...
+    """Backend that manipulates Entries in-memory, for testing purposes."""
+    def __init__(self, data: List[Entry]):
+        self.data = data
+
+    def get_entry(self, uuid: UUID) -> Entry:
+        stack = self.data.copy()
+        while len(stack) > 0:
+            entry = stack.pop()
```
ZLLentz, line 19: Two things here:
- For the existing test self.data, which is a list of one element, this get_entry function essentially does not work, as it would only be able to find the root collection item.
- Can we use pythonic iteration here instead of copy and pop?

```
for entry in stack
```

shilorigins: I'm not sure what you mean.  If an entry isn't the one being searched for, its children are added to the stack, so this function does find entries under the root collection.

Iterating over `self.data` directly would be a problem because we need to track nested entries over the course of the DFS.

superscore/backends/test.py
shilorigins: Added

superscore/backends/test.py
ZLLentz: Am I misunderstanding the code? `self.data` here is being set to a `list` that contains one element.

ZLLentz:
```
    root_col = Collection(
        uuid="441ff79f-4948-480e-9646-55a1462a5a70",
        description="All three facilities in the SLAC LINAC: LCLS-NC, FACET, and LCLS-SC",
        title="Accelerator Directorate",
        children=[
            lcls_nc_col,
            facet_col,
            lcls_sc_col,
        ]
    )

    return TestBackend([root_col])
```

shilorigins:
How do we fix this `Install boa for mambabuild` error?

superscore/backends/test.py
shilorigins:
```
1 def get_entry(self, uuid: UUID) -> Entry:
2         stack = self.data.copy()
3         while len(stack) > 0:
4             entry = stack.pop()
5             if entry.uuid == uuid:
6                 return entry
7             if isinstance(entry, Collection) or isinstance(entry, Snapshot):
8                 stack.extend(entry.children)
9         return None
```
Right, in line 1 stack is a shallow copy of `[root_col]`.  `root_col` is popped off the stack, but it isn't the entry we're looking for.  Since `root_col` is a `Collection`, all entries in `root_col.children` are added to the stack, making inner entries discoverable.

tangkong:
> How do we fix this `Install boa for mambabuild` error?

It's not a required test, so we can still merge once approvals are in.  Unfortunately github thinks optional tests failing = ❌ , to the chagrin of almost everyone.  (Long standing issue)

Aside from this zach and I are looking at the test matrix, but ignore it for now

superscore/backends/test.py
ZLLentz: Oh my god I reviewed too many PRs today, this is literally the sixth one

ZLLentz: sorry

superscore/backends/test.py
```diff
@@ -0,0 +1,49 @@
...
+            if entry.uuid == uuid:
+                return entry
+            if isinstance(entry, Collection) or isinstance(entry, Snapshot):
+                stack.extend(entry.children)
+        return None
```
ZLLentz, line 30: Maybe an actual real question: did we decide on how to handle missing entries in the general case? Should we raise an exception here? We certainly don't need to explicitly `return None`

tangkong: in my other PR I've been returning None, so I think that's my vote

ZLLentz: I think either way is fine- code that calls this simply needs to know what to expect. So we can either raise an exception right away to force calling code to handle it properly (or perish), or we can return None to avoid an automatic perish but expect the caller to check and handle it themselves, else they get some sort of "Nonetype does not have attribute blah" sort of error.

shilorigins: I like how raising an exception parallels the other methods and I think I prefer the automatic perish, but I'm fine with either.

Looking at #25, `FilestoreBackend.get_entry()` uses `dict.get()`, which raises an error if the UUID isn't in the cache. Where you referring to returning `None` somewhere else?

ZLLentz: I think I also prefer the exception raising, if only because "Nonetype cannot do thing" errors are always a bit annoying to handle, versus "I have to catch the raised SuperScoreNoEntry exception"

tangkong: uuh apparently I'm just hallucinating.  Or your future arguments convinced past me 🤷

shilorigins:
`save_entry` and `get_entry` had a mildly ugly interaction when using `ValueError`s, so I've introduced custom exceptions to clean it up.

ZLLentz (approved):
I'm fully on board here

ZLLentz:
CI is a bit borked right now wrt the docs builds so we can ignore that one

tangkong (approved):
Looks good to me too.  Let's get this in

shilorigins:
I can't see how to override a required check. Could one of you merge it in?

tangkong:
You can blame github enterprise for this snafu

PR #28, opened 2024-05-14
Add Entry validation methods

### LINKED ISSUE #11: Build Entry validation logic
- no circular `Collection` and `Snapshot` references
- type hints

shilorigins:
## Description

* Add `.validate()` method for all `Entry` subclasses
* Introduce `Nestable` interface for `Collection` and `Snapshot`
## Motivation and Context

Validation ensures that `Entry`s are well formed before certain uses, such as inserting into a backend: close #11 

Adding validation methods to each `Entry` subclass leverages polymorphism to keep validation code close to the data being validated.  This also keeps the client or backend agnostic to the type of `Entry` being validated.

`Entry.validate()` uses an `apischema` serialization roundtrip to perform type checks. This implementation captures `ValidationError`s and returns `False` if type checks fail, but I'd be open to leaving the exception if that's preferred.  If a subclass needs specific validation tests, it can override this method; the subclass is responsible for calling `Entry.validate()` or otherwise performing type checks.

`Nestable` reduces code duplication between `Collection` and `Snapshot`, and encourages more generalized code when working with these types.
## How Has This Been Tested?

Ran `pytest` from within the repo. Used `linac_backend` and new `linac_snapshot` fixtures to validate the original trees, a tree with an empty `Collection`, a tree with a cycle, and class instances that fail type checks.

<details>
<summary>Test coverage</summary>

```
Name                    Stmts   Miss  Cover
-------------------------------------------
__init__.py                 2      0   100%
__main__.py                 2      2     0%
_version.py                11     11     0%
backends/__init__.py        0      0   100%
backends/core.py           14      5    64%
backends/filestore.py     147     22    85%
backends/test.py           36      0   100%
bin/__init__.py             2      2     0%
bin/help.py                11     11     0%
bin/main.py                51     51     0%
bin/ui.py                  13     13     0%
client.py                  22     22     0%
errors.py                   6      0   100%
model.py                  153      6    96%
serialization.py           61     19    69%
tests/__init__.py           0      0   100%
tests/conftest.py         107      0   100%
tests/test_backend.py      71      0   100%
tests/test_model.py        79      0   100%
tests/test_window.py        5      0   100%
type_hints.py               2      0   100%
utils.py                   11      4    64%
version.py                 27     17    37%
widgets/__init__.py         0      0   100%
widgets/core.py             5      0   100%
widgets/window.py           5      0   100%
-------------------------------------------
TOTAL                     843    185    78%
```
</details>

<details>
<summary>Test durations</summary>

```
0.05s call     superscore/tests/test_window.py::test_main_window
0.04s setup    superscore/tests/test_window.py::test_main_window
0.04s call     superscore/tests/test_backend.py::test_save_entry[0]
0.02s call     superscore/tests/test_model.py::TestEntryValidation::test_fixture_validation
0.01s call     superscore/tests/test_model.py::test_serialize_snapshot_roundtrip
0.01s call     superscore/tests/test_model.py::test_serialize_collection_roundtrip
0.01s call     superscore/tests/test_backend.py::test_search_entry[0]

(47 durations < 0.005s hidden.  Use -vv to show these durations.)
```
</details>

This PR

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/28-entry_validation.rst
+++ b/docs/source/upcoming_release_notes/28-entry_validation.rst
@@ -1,0 +1,23 @@
+28 entry validation
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- Add validation methods to Entry and all subclasses
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- Add Snapshot test fixture mirroring the LINAC Collection
+- Introduce Nestable class interface
+
+Contributors
+------------
+- shilorigins
--- a/superscore/model.py
+++ b/superscore/model.py
@@ -195,16 +195,15 @@ class Nestable:
 
     def has_cycle(self, parents=None) -> bool:
         if parents is None:
-            parents = set()
+            parents = frozenset()
 
         if self.uuid in parents:
             return True
         else:
-            parents.add(self.uuid)
+            parents_including_self = parents | {self.uuid}
             for child in self.children:
-                if isinstance(child, type(self)) and child.has_cycle(parents=parents):
+                if isinstance(child, type(self)) and child.has_cycle(parents=parents_including_self):
                     return True
-            parents.remove(self.uuid)
             return False
 
 
--- a/superscore/model.py
+++ b/superscore/model.py
@@ -9,6 +9,8 @@ from enum import Flag, IntEnum, auto
 from typing import ClassVar, List, Optional, Set, Union
 from uuid import UUID, uuid4
 
+import apischema
+
 from superscore.serialization import as_tagged_union
 from superscore.type_hints import AnyEpicsType
 from superscore.utils import utcnow
@@ -73,6 +75,23 @@ class Entry:
         """
         return []
 
+    def validate(self, toplevel: bool = True) -> bool:
+        """
+        Ensures Entry is properly formed.
+
+        Raises apischema.ValidationError if type checking fails during
+        deserialization
+        """
+        if toplevel:
+            try:
+                serial = apischema.serialize(self)
+                apischema.deserialize(type(self), serial)
+                return True
+            except apischema.ValidationError:
+                return False
+        else:
+            return True
+
 
 @dataclass
 class Parameter(Entry):
@@ -82,6 +101,10 @@ class Parameter(Entry):
     rel_tolerance: Optional[float] = None
     readback: Optional[Parameter] = None
     read_only: bool = False
+
+    def validate(self, toplevel: bool = True) -> bool:
+        readback_is_valid = self.readback is None or self.readback.validate(toplevel=False)
+        return readback_is_valid and super().validate(toplevel=toplevel)
 
 
 @dataclass
@@ -109,6 +132,10 @@ class Setpoint(Entry):
             severity=severity,
             readback=origin.readback,
         )
+
+    def validate(self, toplevel: bool = True) -> bool:
+        readback_is_valid = self.readback is None or self.readback.validate(toplevel=False)
+        return readback_is_valid and super().validate(toplevel=toplevel)
 
 
 @dataclass
@@ -152,8 +179,37 @@ class Readback(Entry):
         )
 
 
+class Nestable:
+    """Mix-in class that provides methods for nested container Entries"""
+    def validate(self, toplevel: bool = True):
+        """
+        Validates self and all children. If toplevel, also validates structure
+        of the Entry tree. This avoids redundant work by only performing tree-
+        level validation once.
+
+        Overrides Entry.validate().
+        """
+        tree_is_valid = not toplevel or (not self.has_cycle() and super().validate(toplevel=True))
+        not_empty = len(self.children) > 0
+        return tree_is_valid and not_empty and all(child.validate(toplevel=False) for child in self.children)
+
+    def has_cycle(self, parents=None) -> bool:
+        if parents is None:
+            parents = set()
+
+        if self.uuid in parents:
+            return True
+        else:
+            parents.add(self.uuid)
+            for child in self.children:
+                if isinstance(child, type(self)) and child.has_cycle(parents=parents):
+                    return True
+            parents.remove(self.uuid)
+            return False
+
+
 @dataclass
-class Collection(Entry):
+class Collection(Nestable, Entry):
     """Nestable group of Parameters and Collections"""
     meta_pvs: ClassVar[List[Parameter]] = []
     all_tags: ClassVar[Set[Tag]] = set()
@@ -180,7 +236,7 @@ class Collection(Entry):
 
 
 @dataclass
-class Snapshot(Entry):
+class Snapshot(Nestable, Entry):
     """
     Nestable group of Values and Snapshots.  Effectively a data-filled Collection
     """
--- a/superscore/model.py
+++ b/superscore/model.py
@@ -190,8 +190,7 @@ class Nestable:
         Overrides Entry.validate().
         """
         tree_is_valid = not toplevel or (not self.has_cycle() and super().validate(toplevel=True))
-        not_empty = len(self.children) > 0
-        return tree_is_valid and not_empty and all(child.validate(toplevel=False) for child in self.children)
+        return tree_is_valid and all(child.validate(toplevel=False) for child in self.children)
 
     def has_cycle(self, parents=None) -> bool:
         if parents is None:
--- a/superscore/tests/test_model.py
+++ b/superscore/tests/test_model.py
@@ -88,6 +88,10 @@ class TestEntryValidation:
         vac_bsy_col.children.append(bsy_col)
         assert not linac_model.validate()
 
+        # ensure there's no hysteresis in cycle detection
+        vac_bsy_col.children.remove(bsy_col)
+        assert linac_model.validate()
+
     @staticmethod
     def test_collection_reachable_from_snapshot_validation(linac_backend):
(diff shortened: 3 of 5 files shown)


docs/source/upcoming_release_notes/28-entry_validation.rst
```diff
@@ -0,0 +1,23 @@
...
+- N/A
+
+Maintenance
+-----------
+- Add Snapshot test fixture mirroring the LINAC Collection
+- Introduce Nestable class interface
+
+Contributors
+------------
+- devagr
```
tangkong, line 23:
```suggestion
- shilorigins
```

We usually use gh usernames here, for better linking in github and external to SLAC

superscore/tests/test_model.py
```diff
@@ -42,3 +42,56 @@ def test_sample_database_roundtrip(sample_database: Root):
     ser = apischema.serialize(Root, sample_database)
     deser = apischema.deserialize(Root, ser)
     assert deser == sample_database
+
+
+class TestEntryValidation:
+    @staticmethod
```
tangkong, line 48: Maybe I've just been in pytest land for too long, but what's the point of putting these tests in a class if they're all `staticmethod`s?

superscore/tests/test_model.py
```diff
@@ -42,3 +42,56 @@ def test_sample_database_roundtrip(sample_database: Root):
...
+    def test_fixture_validation(linac_backend):
+        linac_model = linac_backend.get_entry("441ff79f-4948-480e-9646-55a1462a5a70")
+        assert linac_model.validate()
+
+        linac_snapshot = linac_backend.get_entry("06282731-33ea-4270-ba14-098872e627dc")
+        assert linac_snapshot.validate()
+
+    @staticmethod
+    def test_parameter_readback_validation(linac_backend):
+        vac_gunb_pv1 = linac_backend.get_entry("8f3ac401-68f8-4def-b65a-3c8116c80ba7")
```
tangkong, line 58: I think in general we shouldn't be involving a backend when we test these validation methods.  It should be easy enough to nest fixtures so the `linac_backend` takes a `linac_data` fixture` (which might return, say a `Root`) and produces a backend.  That way we can just access the model dataclasses when we want to test the model dataclasses.

superscore/model.py
```diff
@@ -152,8 +176,39 @@ def from_parameter(
...
+        """
+        return not self.has_cycle() and super().validate() and self.validate_nodes()
+
+    def validate_nodes(self):
+        """
+        Validates self and all children at the Entry level. See Nestable.validate() for
+        more information.
+        """
+        not_empty = len(self.children) > 0
+        all_children_are_valid = all(child.validate_nodes() if isinstance(child, Nestable) else child.validate() for child in self.children)
```
tangkong, line 195: So this calls `child.validate()` for each child, which will call each child's `super.validate()` (`Entry.validate()`) correct?  Would we then end up trying the serialization round trip at every leaf (or every node actually) in the tree?  Despite having also done the round-trip at the top of the tree?

superscore/model.py
```diff
@@ -152,8 +176,39 @@ def from_parameter(
...
+
+    def has_cycle(self, parents=set()) -> bool:
+        if self.uuid in parents:
+            return True
+        else:
+            parents.add(self.uuid)
+            for child in self.children:
+                if isinstance(child, type(self)) and child.has_cycle(parents=parents):
+                    return True
+            parents.remove(self.uuid)
```
tangkong, line 206: If the method terminates here why do we need to remove this uuid?  Is it because of the persistently mutating argument above?

superscore/model.py
```diff
@@ -152,8 +176,39 @@ def from_parameter(
...
+    def validate_nodes(self):
+        """
+        Validates self and all children at the Entry level. See Nestable.validate() for
+        more information.
+        """
+        not_empty = len(self.children) > 0
+        all_children_are_valid = all(child.validate_nodes() if isinstance(child, Nestable) else child.validate() for child in self.children)
+        return not_empty and all_children_are_valid
+
+    def has_cycle(self, parents=set()) -> bool:
```
tangkong, line 198: You should never supply a mutable default argument in python.  The short explanation is that the default arguments are created when the function / method is created, and will be mutated every time you call the method:
```python
In [12]: def has_cycle(parent=set()):
    ...:     parent.add(uuid.uuid4())
    ...:     print(parent)
    ...:

In [13]: has_cycle()
{UUID('ad05302f-e37d-4082-b085-02a011e6fe49')}

In [14]: has_cycle()
{UUID('ad05302f-e37d-4082-b085-02a011e6fe49'), UUID('eed9256a-c83f-4736-9c0d-0249cb0c38d1')}

In [15]: has_cycle()
{UUID('ad05302f-e37d-4082-b085-02a011e6fe49'), UUID('c5eb3967-cbdb-4323-90e2-2293a460a49e'), UUID('eed9256a-c83f-4736-9c0d-0249cb0c38d1')}
```

Instead, you should make the default value `None`, and create a new set inside the method body

tangkong: A more verbose explanation of mutable default arugments https://web.archive.org/web/20200221224620id_/http://effbot.org/zone/default-values.htm

superscore/tests/test_model.py
shilorigins: Just for organization. I ended up leaving "validation" in the test names, but we could shorten them since they're under this validation group. This also lets you run a group of tests using `pytest {file}::{class}`.

docs/source/upcoming_release_notes/28-entry_validation.rst
shilorigins: 🤦‍♂️

superscore/model.py
shilorigins: I think seralization would run once for the whole tree and then once per leaf. Within this function, `child.validate()` is only run for non-`Nestable` entries.

To avoid this time hit, we could modify `Parameter`, `Setpoint`, and `Readback` to distinguish between validation calls as root and as a child, similar to how `Nestable` has two validation functions.

superscore/tests/test_model.py
shilorigins: These tests rely on manipulating the `Entry`s within the fixture (which I prefer over defining a new fixture for every test). If we wanted the fixture to provide a `Root`, we would still need a way to easily access `Entry`s within it. The `TestBackend`'s only purpose is to facilitate this manipulation, and I don't think it gets in the way of the actual test organizationally or functionally.

superscore/model.py
tangkong: `parents` is passed to the children's `has_cycle` calls, but the way you're modifying `parents` **after** those calls are finished implies that `parents` is bleeding out of the scope of this call.

superscore/tests/test_model.py
tangkong: My point here is that it's bad test isolation, and creates situations where changes to the `TestBackend` could cause entry validation tests to also fail.

superscore/model.py
shilorigins: Right, `parents` in this call is the same as `parents` in the parent call, so its scope is larger than this call.

I'll change it so that each node uses a new set.

superscore/model.py
shilorigins: Fixed, plus I accounted for this in the cycle detection test.

superscore/tests/test_model.py
shilorigins: Since I think of `TestBackend` as test infrastructure, I guess it made sense to me that changes to how it works could cause a variety of tests to fail. Regardless, I've refactored these tests to be self-contained, and we can save `linac_backend` for integration tests (if it's still useful then).

superscore/model.py
```diff
@@ -152,8 +176,41 @@ def from_parameter(
         )
 
 
+class Nestable:
+    """Abstract class that provides methods for nested container Entries"""
```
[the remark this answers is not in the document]
shilorigins: Abstract in the sense that it isn't meant to be instantiated, but I take your point 😄

superscore/model.py
```diff
@@ -152,8 +176,41 @@ def from_parameter(
...
+        tree-level validation once.  Overrides Entry.validate().
+        """
+        return not self.has_cycle() and super().validate() and self.validate_nodes()
+
+    def validate_nodes(self):
+        """
+        Validates self and all children at the Entry level. See Nestable.validate() for
+        more information.
+        """
+        not_empty = len(self.children) > 0
```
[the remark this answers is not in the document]
shilorigins: I wasn't considering this as a use case. Since there'd be nothing to restore, I'm not sure how such a snapshot would be useful. If you can envision a user wanting to do this, I can remove the check.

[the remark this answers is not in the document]
tangkong: I think using superscore to simply take historical snapshots of the state of the controls system is a very real use case.  Even if we don't initially intend for this to be the case, if we make saving PV data convenient/easy, users will surely end up taking snapshots freely with no intention to restore all of them.

I guess I always saw the "Save and Restore" as two separate actions the service could perform, not as one compound action.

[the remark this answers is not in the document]
shilorigins: Updated

superscore/model.py
shilorigins: Updated

ZLLentz:
I'm here a little late but everything here looks solid to me
I'll skip hitting approve myself because I think Robert's approval is needed following his previous review

tangkong (approved):
LGTM

PR #34, opened 2024-06-11
TST: Implement test IOC

### LINKED ISSUE #31: Setup caproto test IOC
## Current Behavior

No way to interact with EPICS when writing tests
## Expected Behavior

Ability to read from and write to test-scoped PVs using regular EPICS API, for testing purposes
## Suggested Solution

Setup a pytest fixture that initializes a caproto IOC with PVs from a backend fixture or other specification method

shilorigins:
## Description

Add a fixture and module for instantiating IOCs that make PVs accessible via EPICS during tests and for live interaction (closes #31) 

Fix a bug where the `dummy_cl` fixture was altering the `SHIMS` global variable, causing the IOC test to fail only if run after `test_cl.py`.  I did this both by making a new dict in `dummy_cl`, and by copying `SHIMS` in `ControlLayer.__init__` rather than assigning it to an instance attribute.
### Motivation

This application needs to interface with EPICS via Channel Access and PV Access, so we need a way to test this interface. While mocking can be useful, using actual EPICS tools in at least some tests will make the test suite more complete. This test IOC system is intended to provide test PVs that use the same interface as and behave like production PVs, while being predicable, repeatable, and configurable.

### Design

Because `caproto` IOCs use class attributes to store their PVs, all desired PVs must be passed in before the IOC class is defined, instantiated, and run (there may be a way to reconfigure IOCs after subclassing / instantiation, but I haven't figured it out). `IOCFactory` collects any "PV"s within a group of `Entry` trees, which includes `Parameter`s, `Setpoint`s, and `Readback`s, then defines and instantiates a test IOC subclass. These instances can be used as context managers, automatically starting and stopping themselves for the test scope.

The prefix in the test example helps prevent test PVs from colliding with production PVs in the same EPICS namespace.

### Intended use

This IOC fixture is intended to be used for integration tests, where we can scope the fixture to the module and limit the number of IOC instantiations.

To run the IOC for live interaction, run `python -m superscore.tests.ioc.linac` in a terminal.  You can then query the IOC's PVs from another terminal, or background the process and query it from the same terminal.

## How Has This Been Tested?

Basic PV interaction test in `test_ioc.py`

This PR

## Pre-merge checklist

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,6 +1,8 @@
 # These are required for developing the package (running the tests) but not
 # necessarily required for _using_ it.
+caproto
 coverage
+numpy
 pytest
 pytest-asyncio
 pytest-cov
--- a/superscore/control_layers/core.py
+++ b/superscore/control_layers/core.py
@@ -32,7 +32,7 @@ class ControlLayer:
     def __init__(self, *args, shims: Optional[List[str]] = None, **kwargs):
         if shims is None:
             # load all available shims
-            self.shims = SHIMS
+            self.shims = SHIMS.copy()
             logger.debug('No shims specified, loading all available communication '
                          f'shims: {list(self.shims.keys())}')
         else:
--- a/superscore/tests/conftest.py
+++ b/superscore/tests/conftest.py
@@ -742,8 +742,7 @@ class DummyShim(_BaseShim):
 @pytest.fixture(scope='function')
 def dummy_cl() -> ControlLayer:
     cl = ControlLayer()
-    cl.shims['ca'] = DummyShim()
-    cl.shims['pva'] = DummyShim()
+    cl.shims = {protocol: DummyShim() for protocol in ['ca', 'pva']}
     return cl
 
 
--- a/superscore/tests/conftest.py
+++ b/superscore/tests/conftest.py
@@ -13,10 +13,10 @@ from superscore.control_layers._base_shim import _BaseShim
 from superscore.control_layers.core import ControlLayer
 from superscore.model import (Collection, Parameter, Readback, Root, Setpoint,
                               Snapshot)
+from superscore.tests.ioc import IOCFactory
 
 
-@pytest.fixture(scope='function')
-def linac_backend():
+def linac_data():
     lasr_gunb_pv1 = Parameter(
         uuid="5544c58f-88b6-40aa-9076-f180a44908f5",
         pv_name="LASR:GUNB:TEST1",
@@ -615,6 +615,12 @@ def linac_backend():
         ]
     )
 
+    return all_col, all_snapshot
+
+
+@pytest.fixture(scope='function')
+def linac_backend():
+    all_col, all_snapshot = linac_data()
     return TestBackend([all_col, all_snapshot])
 
 
@@ -772,3 +778,10 @@ def sample_client(
     client.cl = dummy_cl
 
     return client
+
+
+@pytest.fixture(scope='module')
+def linac_ioc():
+    _, snapshot = linac_data()
+    with IOCFactory.from_entries(snapshot.children)(prefix="SCORETEST:") as ioc:
+        yield ioc
--- a/superscore/tests/ioc/ioc_factory.py
+++ b/superscore/tests/ioc/ioc_factory.py
@@ -1,0 +1,71 @@
+from multiprocessing import Process
+from typing import Iterable, Mapping, Union
+
+from caproto.server import PVGroup, pvproperty
+from caproto.server import run as run_ioc
+from epicscorelibs.ca import dbr
+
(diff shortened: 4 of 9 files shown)


tangkong:
Is there still work being done on this?

shilorigins:
I haven't decided how to set up the actual fixture. The code here works though, so I can touch it up and submit for review if we don't mind adding the fixture in a later PR.

tangkong:
A suggestion that might result in a refactor, but might also fix our segfault issues:

Maybe we should make explicit modules that hold these iocs, similar to `caproto.ioc_examples.random_walk` etc.  This would allow us to have people try superscore against the GUI outside of the test suite.  Critically, this would also let us use the test-ioc pattern used in caproto: [run_example_ioc](https://github.com/caproto/caproto/blob/46020e4816c7dbe8ccae9b7f0af362ce9a7f4f07/caproto/tests/conftest.py#L59)

Maybe a hook in pytest would regenerate those module files based on the fixtures we've defined as well

dev-requirements.txt
```diff
@@ -1,6 +1,8 @@
 # These are required for developing the package (running the tests) but not
 # necessarily required for _using_ it.
+caproto
```
tangkong, line 3: we also need to add these requirements to the `[meta.yml](https://github.com/pcdshub/superscore/blob/dc7ded7f122acec96393421ffc8b7efb4a514391/conda-recipe/meta.yaml#L32-L36)` file's test requirements to get the conda builds passsing

shilorigins:
`caproto` keeps fighting me, but I've gotten it to the point where we can launch the IOC and manually query it, strings don't get mapped to char arrays, and the test passes when run individually (`pytest tests/test_ioc.py`).  However, the test fails when the whole suite is run (`pytest`).  Once I figure out the discrepancy, I still need to clean up the code as well.

shilorigins:
I resolved the individual vs suite testing issue (`dummy_cl` fixture was altering the `SHIMS` global variable). There's some work to be done making the module and IOC fixture more interdependent, but this is enough for one PR.  I only included basic tests for the IOC itself, not integration tests.

ZLLentz (approved):
I think this looks really nice! Good job finding the SHIMS issue and winning the fight.

superscore/tests/ioc/ioc_factory.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        Defines and instantiates a TempIOC subclass containing all PVs reachable
+        from entries.
+        """
+        attrs = IOCFactory.prepare_attrs(entries)
+        IOC = type("IOC", (TempIOC,), attrs)
+        return IOC
+
+    @staticmethod
+    def collect_pvs(entries: Iterable[Entry]) -> Iterable[Union[Parameter, Setpoint, Readback]]:
```
tangkong, line 45: We've written this function a couple of times at this point (or at least similar looking ones), we should find a good way to DRY this out (later, not this PR)

tangkong (approved):
A hard fought, and pretty awesome bit of test suite functionality.  I just had a couple of questions, but nothing that needs to change.  Nice work! 👍

tangkong:
@shilorigins Any last thoughts or can we merge this?

superscore/control_layers/core.py
```diff
@@ -32,7 +32,7 @@ class ControlLayer:
     def __init__(self, *args, shims: Optional[List[str]] = None, **kwargs):
         if shims is None:
             # load all available shims
-            self.shims = SHIMS
+            self.shims = SHIMS.copy()
```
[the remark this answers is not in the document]
shilorigins: This change is redundant for fixing the issue, but I figured it would help avoid similar mistakes in the future.  If we want to configure shims for a session, as you say, we can revert this line later.

superscore/tests/conftest.py
```diff
@@ -13,10 +13,10 @@
 from superscore.control_layers.core import ControlLayer
 from superscore.model import (Collection, Parameter, Readback, Root, Setpoint,
                               Snapshot)
+from superscore.tests.ioc import IOCFactory
 
 
-@pytest.fixture(scope='function')
-def linac_backend():
+def linac_data():
```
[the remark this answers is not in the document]
shilorigins: Marking it as a fixture makes it difficult to use outside of a `pytest` context, which I ran into when trying to run the test IOC. You can still import and call it in other fixtures, it just won't happen automatically.  If we do want fixture controls, I suppose we could make a fixture that simply calls and returns `linac_data` to get the benefits of both options.

superscore/tests/ioc/ioc_factory.py
shilorigins: I agree.  As a note for the future PR, the reason I didn't do it here is because it requires untangling backend calls in the client version, which are required for resolving UUIDs.

PR #35, opened 2024-06-12
ENH: Control layer and shim layers

tangkong:
## Description
- adds Shim layers, for unifying interface between different EPICS communication libraries
- adds ControlLayer, which dispatches to the appropriate shim layer.  The intent is for the client to hold onto a `ControlLayer` and send requests to the control system through it.

### Remaining to-do, misc thoughts
- [x] Sketch out unified "status" to track status / success of actions
- [x] Determine if shim layers need to be singletons  (No, this is fine)
- [x] Figure out what to do with data-types within shims (as is aioca only uses datatype to coerce CA information.  ignore for now)
- [x] clean up timeout errors and other dso debug output (These are log-level related.  Asyncio and setuptools_dso both have debug output)

## Motivation and Context (read: stream of tangkong's consciousness)
<details><summary>Async thoughts</summary>
<p>

### Async thoughts
We have the opportunity here to be more async about this from the start.  Since we're doing a lot of effectively IO-bound tasks (sending signal to EPICS and waiting for a response), async was a good fit.  However, I did not want to expose the async interface to the user.  I think it makes sense to stop the async-ing at the ControlLayer, so that the Client (and user) won't be exposed.  This is important if we want, frankly, anybody else at SLAC to use this code. 

#### Task Cancelling
`asyncio` has a concept of task cancelling, in which the Task receives a cancel signal and raises a CancelledError.  After this the task still needs to be awaited to complete.  Critically this is something that happens in an async coroutine.  In our case, we're sync-wrapping everything at the control layer level, so users won't have the opportunity to do this.  As such I've decided not to wrestle with this asyncio feature, though in the future we might. 

</p>
</details> 

<details><summary>What's this weird `TaskStatus` thing?</summary>
<p>

Each communication library (p4p, pyepics, aioca, etc) handles status / timeout / connection errors a bit differently.  We need a unified way to handle this in the `ControlLayer`, and for this reason I took inspiration from how `ophyd` and `ophyd-async` handle long processes.  This also gives us a place to run single-shot callbacks, since these communication libraries typically only expose a `camonitor` method, which runs callbacks on each update.  

</p>
</details> 

<details><summary>.pyi stub files</summary>
<p>

### .pyi what?
These are stub files used for static type checkers such such as mypy and pylance.  Normally we don't need this but I was trying to make `@singledispatchmethod` work.  Turns out python evaluates `@singledispatch` type decorators at runtime, using some magic shenanigans that static type checkers don't pick up.  To help people IDE's, stub files can be used with `@overload` to hint at these dispatched signatures.  For me this renders quite nicely, showing both signatures and the docstring of the application-code method.

image

Critically, overload and single-dispatch don't seem to play nicely together.  It is possible to define all these without stub files, but I thought moving that boiler plate out cleaned things up (and eliminated the need for `TYPE_CHECKING` check blocks)

</p>
</details> 

## How Has This Been Tested?
Interactively mostly, tests to come.

This PR

## Pre-merge checklist

### DIFF
--- a/superscore/control_layers/_base_shim.py
+++ b/superscore/control_layers/_base_shim.py
@@ -1,3 +1,6 @@
+"""
+Base shim abstract base class
+"""
 from typing import Any, Callable
 
 
--- a/superscore/control_layers/_base_shim.py
+++ b/superscore/control_layers/_base_shim.py
@@ -1,0 +1,12 @@
+from typing import Any, Callable
+
+
+class _ShimBase:
+    async def get(self, address: str):
+        raise NotImplementedError
+
+    async def put(self, address: str, value: Any):
+        raise NotImplementedError
+
+    def monitor(self, address: str, callback: Callable):
+        raise NotImplementedError
--- a/superscore/control_layers/_base_shim.py
+++ b/superscore/control_layers/_base_shim.py
@@ -3,9 +3,11 @@ Base shim abstract base class
 """
 from typing import Any, Callable
 
+from superscore.type_hints import AnyEpicsType
+
 
 class _BaseShim:
-    async def get(self, address: str):
+    async def get(self, address: str) -> AnyEpicsType:
         raise NotImplementedError
 
     async def put(self, address: str, value: Any):
--- a/superscore/control_layers/_base_shim.py
+++ b/superscore/control_layers/_base_shim.py
@@ -1,7 +1,7 @@
 from typing import Any, Callable
 
 
-class _ShimBase:
+class _BaseShim:
     async def get(self, address: str):
         raise NotImplementedError
 
--- a/superscore/control_layers/core.py
+++ b/superscore/control_layers/core.py
@@ -4,7 +4,9 @@ and dispatches to various shims depending on the context.
 """
 import asyncio
 from functools import singledispatchmethod
-from typing import Any, Dict
+from typing import Any, Callable, Dict, Optional
+
+from superscore.control_layers.status import TaskStatus
 
 from ._aioca import AiocaShim
 from ._base_shim import _ShimBase
@@ -59,26 +61,46 @@ class ControlLayer:
         return await shim.get(pv)
 
     @singledispatchmethod
-    def put(self, pv, value: Any):
+    def put(self, pv, value: Any, cb: Optional[Callable]):
         print(f"PV is of an unsupported type: {type(pv)}. Provide either "
               "a string or list of strings")
 
     @put.register
-    def _(self, pv: str, value: Any):
-        return asyncio.run(self._put_one(pv, value))
+    def _(self, pv: str, value: Any, cb: Optional[Callable] = None):
+        async def status_coro():
+            status = self._put_one(pv, value)
+            if cb is not None:
+                status.add_callback(cb)
+            await status.task
+            return status
+
+        return asyncio.run(status_coro())
 
     @put.register
-    def _(self, pv: list, value: list, sequential: bool = False):
-        coros = []
-        for p, val in zip(pv, value):
-            coros.append(self._put_one(p, val))
+    def _(self, pv: list, value: list, cb: Optional[list[Callable]] = None):
 
-        loop = asyncio.get_event_loop()
-        return loop.run_until_complete(asyncio.gather(*coros))
+        async def status_coros():
+            statuses = []
+            if cb is None:
+                callbacks = [None for _ in range(len(pv))]
+            else:
+                callbacks = cb
 
+            for p, val, c in zip(pv, value, callbacks):
+                status = self._put_one(p, val)
+                if c is not None:
+                    status.add_callback(c)
+
+                statuses.append(status)
+            await asyncio.gather(*[s.task for s in statuses])
+            return statuses
+
+        return asyncio.run(status_coros())
+
+    @TaskStatus.wrap
     async def _put_one(self, pv: str, value: Any):
         shim = self.shim_from_pv(pv)
-        return await shim.put(pv, value)
+        await shim.put(pv, value)
 
     def subscribe(self, pv, cb):
         # Subscribes a callback to the PV address
--- a/superscore/control_layers/core.py
+++ b/superscore/control_layers/core.py
@@ -4,7 +4,7 @@ and dispatches to various shims depending on the context.
 """
 import asyncio
 from functools import singledispatchmethod
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Callable, Dict, Optional, Union
 
 from superscore.control_layers.status import TaskStatus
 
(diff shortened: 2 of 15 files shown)


tangkong:
I could go on but I think this is a good stopping point.  We need to move on to Client level functions and GUI things for imminent internship stuff.

superscore/control_layers/core.py
```diff
@@ -0,0 +1,204 @@
...
+    whichever shim is relevant.
+    """
+    shims: Dict[str, _BaseShim]
+
+    def __init__(self, *args, **kwargs):
+        self.shims = {
+            'ca': AiocaShim(),
+        }
+
+    def shim_from_pv(self, pv: str) -> _BaseShim:
```
ZLLentz, line 27: There's some inconsistency between the shims calling this parameter the "address" and the control layer calling this parameter the "pv". I think these are actually swapped, where this one is expecting an "address" ("pva://things") and the shim is expecting just a "pv" ("things"). Maybe it'd be least confusing if we used the same term on each interface.

superscore/control_layers/core.py
```diff
@@ -0,0 +1,204 @@
...
+        else:
+            # No comms mode specified, use the default
+            shim = list(self.shims.values())[0]
+
+        if shim is None:
+            raise ValueError(f"PV is of an unsupported protocol: {pv}")
+
+        return shim
+
+    @singledispatchmethod
```
ZLLentz, line 61: TIL about this tool

superscore/control_layers/status.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @classmethod
+    def wrap(cls: Type[TS], f: Callable[..., Awaitable]) -> Callable[..., TS]:
+        """Wrap an async function in a TaskStatus."""
+
+        @functools.wraps(f)
+        def wrap_f(*args, **kwargs) -> TS:
+            return cls(f(*args, **kwargs))
+
+        return cast(Callable[..., TS], wrap_f)
```
ZLLentz, line 80: The precise magic of how this works escapes me a bit but I'm glad it works

superscore/tests/test_status.py
ZLLentz, line 1: These tests are clear enough to double nicely as a way to understand how the `TaskStatus` is intended to work and be used

ZLLentz (commented):
I like this, I added some questions and food for thought

superscore/control_layers/core.py
shilorigins: Same, I just was reading about this 🙂

superscore/control_layers/_base_shim.py
```diff
@@ -0,0 +1,15 @@
+"""
+Base shim abstract base class
+"""
+from typing import Any, Callable
+
+
+class _BaseShim:
+    async def get(self, address: str):
```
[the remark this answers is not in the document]
shilorigins: We already have `AnyEpicsType` defined in `type_hints.py`, but it doesn't include lists

superscore/control_layers/status.py
tangkong: This is entirely for type-hinting, if I'm honest I wouldn't have come up with this on my own

superscore/control_layers/core.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        shim = self.shim_from_pv(pv)
+        return await shim.get(pv)
+
+    @singledispatchmethod
+    def put(
+        self,
+        pv: Union[str, list[str]],
+        value: Union[Any, list[Any]],
+        cb: Optional[Callable] = None
```
shilorigins, line 110:
```suggestion
        cb: Optional[Union[Callable, list[Callable]]] = None
```
As-is, the type hint doesn't match the pattern of the other parameters.  Not sure if that actually affects anything

superscore/tests/test_cl.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_put(dummy_cl):
+    result = dummy_cl.put("OTHER:PREFIX", 4)
+    assert isinstance(result, TaskStatus)
+    assert result.done
+
+    results = dummy_cl.put(["OTHER:PREFIX", "GE", "LT"], [4, 5, 6])
+    assert all(isinstance(res, TaskStatus) for res in results)
+    assert result.done
```
shilorigins, line 27: Is this supposed to check whether `result` is still done?  Or should it check whether the `TaskStatus`es in `results` are all done?

superscore/control_layers/_base_shim.py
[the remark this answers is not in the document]
ZLLentz: These epics layers might be sending waveforms as numpy arrays, actually. I remember now that we ignored this in type_hints because settable arrays was not a common use case.

superscore/control_layers/_base_shim.py
```diff
@@ -0,0 +1,15 @@
...
+        raise NotImplementedError
+
+    async def put(self, address: str, value: Any):
+        raise NotImplementedError
+
+    def monitor(self, address: str, callback: Callable):
```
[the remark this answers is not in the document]
tangkong: This is definitely something to add.  I included monitor here for completeness, but sort of envisioned us mostly using the status objects for callback purposes.  I looked into it briefly early on, and while aioca allows you to unsubscribe specific callbacks, pyepics only seems to support unsubscribing an entire monitor.

superscore/control_layers/_base_shim.py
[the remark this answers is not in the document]
tangkong: I do recall making a choice to not support lists.  I think `AnyEpicsType` here is appropriate for us right now.

superscore/control_layers/core.py
tangkong: Nothing in the type-hints affects runtime behavior, it's all for static analysis and our convenience.  My analysis tools look directly at the dispatch methods for suggestions, and not the base dispatcher method.  

That being said I will definitely make this change 👍

superscore/tests/test_cl.py
tangkong: whoops copy-pasta error.  That ends up checking the previous result object, vs plural-`results` which is a list of `TaskStatus` objects

superscore/control_layers/_base_shim.py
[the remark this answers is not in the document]
tangkong: I might make a follow up issue for this.  I want to spend some time looking at how other EPICS layers handle unsubscribing and form the implementation then.

shilorigins (approved):

ZLLentz (approved):

tangkong:
Huh I guess this doesn't have a proper ticket to close.  I'll just merge and move on

PR #38, opened 2024-06-21
GUI: Basic GUI elements, DataWidget pages

tangkong:
## Description
- Adds `DataWidget` and `NameDescTagsWidget`, components for making widgets that communicate with dataclasses
- Adds `EntryItem` and `RootTree` for representing Entry dataclasses in a tree view
- Adds one sample Collection widget to demonstrate how to use these widget mixins

To-Do, possibly in followup issues:
- Make a QDataclassBridge type for a set specifically, to support tag addition
- Rework Tags entry when we figure out where we're storing tags

## Motivation and Context
We need a framework to make a GUI.

I lifted most of this from ATEF, and as such we should be free to challenge any of the things we've built here.  We've found it quite useful, and I think it can be similarly useful here.  There should definitely be a keen eye taken to these.  While we've proven they work, they shouldn't be taken as gospel.  I say this partially for the review of this PR, but maybe mostly for future efforts that use these components.

The `RootModel` is very incomplete, and client interactions still need to be fleshed out.  I can see a world where we create the client on GUI startup and have some helper for fetching it, rather than passing it to every widget instance.   There was also an intent to make the data model loading lazy, and the `RootModel` is where this would manifest. But this PR is already long enough. 

If I ever have to vendor the DataWidget / QDataclassBridge stuff again I'm going to upstream it into pcdsutils.

## How Has This Been Tested?
Added some tests.

This PR

image

## Pre-merge checklist

### DIFF
--- a/superscore/widgets/core.py
+++ b/superscore/widgets/core.py
@@ -1,10 +1,17 @@
 """
-Core classes for qt-based GUIs.
+Core widget classes for qt-based GUIs.
 """
+from __future__ import annotations
+
 from pathlib import Path
+from typing import ClassVar, List
+from weakref import WeakValueDictionary
 
 from pcdsutils.qt.designer_display import DesignerDisplay
+from qtpy import QtWidgets
 
+from superscore.qt_helpers import QDataclassBridge, QDataclassList
+from superscore.type_hints import AnyDataclass
 from superscore.utils import SUPERSCORE_SOURCE_PATH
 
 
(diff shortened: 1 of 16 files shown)


superscore/widgets/core.py
```diff
@@ -1,14 +1,398 @@
...
+    ] = WeakValueDictionary()
+    bridge: QDataclassBridge
+    data: AnyDataclass
+
+    def __init__(self, data: AnyDataclass, **kwargs):
+        super().__init__(**kwargs)
+        self.data = data
+        try:
+            # TODO figure out better way to cache these
+            # TODO worried about strange deallocation timing race conditions
```
shilorigins, line 60: Could we make issues for these tasks? Otherwise we'll probably lose track of them.

superscore/widgets/tree.py
```diff
@@ -0,0 +1,415 @@
...
+
+        if column == 0:
+            if isinstance(self._data, Nestable):
+                return getattr(self._data, 'title', 'root')
+            else:
+                return getattr(self._data, 'pv_name')
+        elif column == 1:
+            return getattr(self._data, 'description', 'fart')
+
+        # TODO: something about icons
```
shilorigins, line 82: Is `fart` our official sentinel string? 😆

tangkong: oh god I forgot about this.

superscore/widgets/core.py
tangkong: We can, though we've not had issues with this specific caching bit in quite some time.  I'll make some issues on the condition that we don't actually work on them soon 😆

tangkong: see #52

shilorigins (approved):
I'm happy with this, thanks!

PR #53, opened 2024-07-09
ENH: add Client.apply method, adjust TaskStatus to properly capture exceptions

### LINKED ISSUE #7: Write apply-Snapshot-to-EPICS routine
aka: write the method that applies a `Snapshot` to EPICS.  The `Client.apply` method

First pass, simply walk through the requested `Snapshot` and apply each `Value` to EPICS.  Optionally wait on associated readback PVs.  

On a second pass, should deal with Client-side authentication (allowed actions).  
This is a client method, and will not involve database authentication (handled by backend?)

# Dependencies
- [ ] #31

tangkong:
## Description
- Adds Client.apply method, for writing values to the control system
- Adjusts use of `TaskStatus` to actually capture exceptions, instead of propagating them

## Motivation and Context
closes #7 

Assorted thoughts
- In the future it may be worth validating these Entries before applying them
- `TaskStatus` was built to hold exceptions raised in the coroutines they wrap, but were propagating them to the calling context.  This was caused by our used of `asyncio.gather`, which needed the `return_exceptions=True` kwarg.  `asyncio` has various ways of handling and propagating exceptions, we just needed to choose the right one for us.

## How Has This Been Tested?
Basic unit test added

This PR, docstrings etc

## Pre-merge checklist

### DIFF
--- a/superscore/client.py
+++ b/superscore/client.py
@@ -1,16 +1,23 @@
 """Client for superscore.  Used for programmatic interactions with superscore"""
-from typing import Any, Generator
+import logging
+from typing import Any, Generator, List, Optional, Union
+from uuid import UUID
 
 from superscore.backends.core import _Backend
-from superscore.model import Entry
+from superscore.control_layers import ControlLayer
+from superscore.control_layers.status import TaskStatus
+from superscore.model import Entry, Setpoint, Snapshot
+
+logger = logging.getLogger(__name__)
 
 
 class Client:
     backend: _Backend
+    cl: ControlLayer
 
-    def __init__(self, backend=None, **kwargs) -> None:
-        # if backend is None, startup default filestore backend
-        return
+    def __init__(self, backend: _Backend, **kwargs) -> None:
+        self.backend = backend
+        self.cl = ControlLayer()
 
     @classmethod
     def from_config(cls, cfg=None):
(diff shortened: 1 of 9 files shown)


superscore/client.py
```diff
@@ -34,9 +41,106 @@ def compare(self, entry_l: Entry, entry_r: Entry) -> Any:
...
+            return [self.cl.put(entry.pv_name, entry.data)]
+
+        # Gather pv-value list and apply at once
+        status_list = []
+        pv_list, data_list = self._gather_data(entry)
+        if sequential:
+            for pv, data in zip(pv_list, data_list):
+                logger.debug(f'Putting {pv} = {data}')
+                status: TaskStatus = self.cl.put(pv, data)
+                if status.exception():
```
ZLLentz, line 82: Is `status` guaranteed to be complete after `self.cl.put` returns it?

superscore/client.py
```diff
@@ -34,9 +41,106 @@ def compare(self, entry_l: Entry, entry_r: Entry) -> Any:
...
+        -------
+        Optional[List[TaskStatus]]
+            TaskStatus(es) for each value applied.
+        """
+        if not isinstance(entry, (Setpoint, Snapshot)):
+            logger.info("Entries must be a Snapshot or Setpoint")
+            return
+
+        if isinstance(entry, Setpoint):
+            return [self.cl.put(entry.pv_name, entry.data)]
```
ZLLentz, line 73: Given the extra wrapping of list/not list here, I have to wonder if it would have been simpler for `put` to always return a list of statuses, even if we only put one PV

superscore/control_layers/status.py
```diff
@@ -57,6 +59,27 @@ def success(self) -> bool:
             and self.task.exception() is None
         )
 
+    def wait(self, timeout=None) -> None:
```
ZLLentz, line 62: Small documentation thing: you've added this `wait` function here and not documented its inclusion in the release notes or in the PR text. It's also not used anywhere outside the test suite yet.

superscore/client.py
tangkong: It does currently, since we `asyncio.run` and block until the coroutine is finished.  I'm not entirely convinced this is the best way to do things, but returning the status immediately would require us either
* be in an async context
* run the event loop in a separate thread (ophyd style).

I've been ok with this for now since we have the ability to run parallel puts (and wait on all of them), but it has been something I've been pondering changing

superscore/control_layers/status.py
tangkong: I thought I'd need it, then I realized the full ramifications of the statuses I'd created 😆

superscore/client.py
ZLLentz: I think this is totally ok, I wanted to make sure you hadn't skip a `wait` call here (since `wait` was added in this PR).
It's not so strange to require more nuanced asynchronous behavior to be done in an async context.

superscore/client.py
tangkong: I had this thought too.  In the end I decided that if I, as a user of the `Client`, put to a single PV and had to index into a single-element list, I'd be annoyed.

But maybe the users and I can just deal with it 🤷

ZLLentz: Either way is probably ok, your explanation here makes sense to me

superscore/client.py
```diff
@@ -34,9 +41,106 @@ def compare(self, entry_l: Entry, entry_r: Entry) -> Any:
...
+        self,
+        entry: Union[Setpoint, Snapshot],
+        sequential: bool = False
+    ) -> Optional[List[TaskStatus]]:
+        """
+        Apply settings found in ``entry``.  If no writable values found, return.
+        If ``sequential`` is True, apply values in ``entry`` in sequence, blocking
+        with each put request.  Else apply all values simultaneously (asynchronously)
+
+        Returns
```
shilorigins, line 54:
```suggestion
```

shilorigins: Just a typo in the docstring

superscore/tests/test_status.py
```diff
@@ -40,7 +51,34 @@ async def test_status_fail(failing_coroutine):
     with pytest.raises(ValueError):
         await status
 
-    assert type(status.exception()) == ValueError
+    assert isinstance(status.exception(), ValueError)
```
shilorigins, line 54: I'm getting lost on when exceptions are propagated and when they're captured.  In `test_cl.py:test_fail` it says `exceptions get passed through the control layer` and `exceptions get captured in status object`, but here the exception propagates even though I don't see `return_exceptions` set to `False` anywhere. Could you explain the design for exception handling?

shilorigins: Looking at this with fresh eyes, the lines in `test_cl.py:test_fail` are referencing that `ControlLayer.get` propagates errors while `ControlLayer.put` captures them.  And the `TaskStatus` in this test raises an error because capturing happens within `ControlLayer.put` (more specifically, `asyncio.gather`), not within `TaskStatus`, which was my point of confusion.

shilorigins (approved):
I'm happy with this!

superscore/tests/test_status.py
tangkong: I don't blame you for being confused, I was clearly confused when I first put this together.  To be clear:
* `asyncio.gather(*tasks, return_exceptions = True)` --> exceptions are stored in the `asyncio.Task`, and by extension the `TaskStatus`.  Tasks that raise exceptions are treated the same way as successful Tasks
* `asyncio.gather(*tasks, return_exceptions = False)` --> The first exception raised is propagated to the task that is awaiting the `gather` call.  This means that the coroutine stack stops execution if an exception is thrown.
  * In our situation, this would prevent the TaskStatus from being returned
```python
def nested_coro():
    status = await asyncio.gather(*tasks, return_exceptions=False)
    # exceptions stop execution here
    return status
```

tangkong: With more specific references to our case:
* `ControlLayer.get()` does not wrap any calls in `TaskStatus`, nor does it `asyncio.gather`.  Any exceptions thrown in the `.get()` method will propagate and halt execution
* `ControlLayer.put()` wraps `._put_one` in a `TaskStatus`.  These statuses are then awaited in `asyncio.gather(*statuses, return_exceptions=True)`. Exceptions raised in `._put_one` will be captured by the `asyncio.Task`, which is examined by `TaskStatus`

tangkong:
Alright, merging and moving along

PR #54, opened 2024-07-11
ENH: Add ability to instantiate Client from configuration file

### LINKED ISSUE #8: Establish configuration file format and loading constructor
## Basic description

Build configuration file loading and parsing mechanism for initializing the Client / GUI.  It's probably worthwhile to keep all of the configuration in one place, and just have the client reject whatever parameters the GUI needs and visa versa.

Should include
* Initialization kwargs for backend layer (db location for filestore, database url and credentials for database, etc)
* Access control configuration
* Allowed tags?

I expect it's most natural for this to be a yaml file, and for the client to take a path to this file as an initialization argument.  The CLI will also need an argument for providing this path, or a way of discovering the path via environment variable.

## Other thoughts

Down the line I expect there to be some admin configuration that takes precedence over users' individual configurations.  That is: we have some global configuration (access groups, set of tags, permitted actions), and a user sets up their session under that umbrella.  It likely makes little sense for this to live in a configuration file, but

tangkong:
## Description
- Adds ability for `Client` to search for configuration files, and load settings from them.
  - Searches, in order:
    - `$SUPERSCORE_CFG` environment variable
    - `$XDG_CONFIG_HOME` directory, for files named `superscore.cfg` or `.superscore.cfg`
    - `~/.config` directory, for files named `superscore.cfg` or `.superscore.cfg`

## Motivation and Context
closes #8 

## How Has This Been Tested?
Added a couple tests.  Otherwise interactive tests

This PR

## Pre-merge checklist

### DIFF
--- a/superscore/backends/__init__.py
+++ b/superscore/backends/__init__.py
@@ -1,0 +1,38 @@
+__all__ = ['BACKENDS']
+
+import logging
+from typing import Dict
+
+from .core import _Backend
+
+logger = logging.getLogger(__name__)
+
+
+def _get_backend(backend: str) -> _Backend:
+    if backend == 'filestore':
+        from .filestore import FilestoreBackend
+        return FilestoreBackend
+    if backend == 'test':
+        from .test import TestBackend
+        return TestBackend
+
+    raise ValueError(f"Unknown backend {backend}")
+
+
+def _get_backends() -> Dict[str, _Backend]:
+    backends = {}
+
+    try:
+        backends['filestore'] = _get_backend('filestore')
+    except ImportError as ex:
+        logger.debug(f"Filestore Backend unavailable: {ex}")
+
+    try:
+        backends['test'] = _get_backend('test')
+    except ImportError as ex:
+        logger.debug(f"Test Backend unavailable: {ex}")
+
+    return backends
+
+
+BACKENDS = _get_backends()
--- a/superscore/backends/__init__.py
+++ b/superscore/backends/__init__.py
@@ -8,6 +8,9 @@ from .core import _Backend
 logger = logging.getLogger(__name__)
 
 
+BACKENDS: Dict[str, _Backend] = {}
+
+
 def _get_backend(backend: str) -> _Backend:
     if backend == 'filestore':
         from .filestore import FilestoreBackend
@@ -19,7 +22,7 @@ def _get_backend(backend: str) -> _Backend:
     raise ValueError(f"Unknown backend {backend}")
 
 
-def _get_backends() -> Dict[str, _Backend]:
+def _init_backends() -> Dict[str, _Backend]:
     backends = {}
 
     try:
@@ -35,4 +38,21 @@ def _get_backends() -> Dict[str, _Backend]:
     return backends
 
 
-BACKENDS = _get_backends()
+def get_backend(backend_name: str) -> _Backend:
+    try:
+        backend = BACKENDS[backend_name]
+    except KeyError:
+        # try to load it
+        try:
+            BACKENDS[backend_name] = _get_backend(backend_name)
+            backend = BACKENDS[backend_name]
+        except ValueError:
+            raise ValueError(f'Backend {backend_name} not supported. Available '
+                             f'backends include: ({list(BACKENDS.keys())})')
+        except ImportError as ex:
+            raise ValueError(f'Backend {(backend_name)} failed to load: {ex}')
+
+    return backend
+
+
+BACKENDS = _init_backends()
--- a/superscore/client.py
+++ b/superscore/client.py
@@ -35,6 +35,39 @@ class Client:
 
     @classmethod
     def from_config(cls, cfg: Path = None):
+        """
+        Create a client from the configuration file specification.
+
+        Configuration file should be of an "ini" format, along the lines of:
+
+        .. code::
+
+            [backend]
+            type = filestore
+            path = ./db/filestore.json
+
+            [control_layer]
+            ca = true
+            pva = true
+
+        The ``backend`` section has one special key ("type"), and the rest of the
+        settings are passed to the appropriate ``_Backend`` as keyword arguments.
+
+        The ``control_layer`` section has a key-value pair for each available shim.
+        The ``ControlLayer`` object will be created with all the valid shims with
+        True values.
+
+        Parameters
+        ----------
+        cfg : Path, optional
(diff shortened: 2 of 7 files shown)


superscore/backends/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+def _get_backend(backend: str) -> _Backend:
+    if backend == 'filestore':
+        from .filestore import FilestoreBackend
+        return FilestoreBackend
+    if backend == 'test':
+        from .test import TestBackend
+        return TestBackend
+
+    raise ValueError(f"Unknown backend {backend}")
```
ZLLentz, line 19: Nitpick: maybe this error message can tell us what the known backends are

superscore/backends/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def _get_backends() -> Dict[str, _Backend]:
+    backends = {}
+
+    try:
+        backends['filestore'] = _get_backend('filestore')
+    except ImportError as ex:
+        logger.debug(f"Filestore Backend unavailable: {ex}")
```
ZLLentz, line 28: I understand the ethos of not falling over here, but doesn't an unimportable backend indicate some sort of major problem? Are we expecting users to maybe not have various backend support modules installed?

superscore/client.py
```diff
@@ -15,13 +19,124 @@ class Client:
...
 
-    def __init__(self, backend: _Backend, **kwargs) -> None:
+    def __init__(
+        self,
+        backend: Optional[_Backend] = None,
+        control_layer: Optional[ControlLayer] = None,
+    ) -> None:
+        if backend is None:
+            # set up a temp backend with temp file
+            backend = BACKENDS['test']
```
ZLLentz, line 29: Given the handling in `_get_backends()`, it's possible this could fail with a cryptic `KeyError`

superscore/client.py
```diff
@@ -15,13 +19,124 @@ class Client:
...
+
+        # Gather Backend
+        if 'backend' in cfg_parser.sections():
+            backend_type = cfg_parser.get("backend", "type")
+            kwargs = {key: value for key, value
+                      in cfg_parser["backend"].items()
+                      if key != "type"}
+            backend = BACKENDS[backend_type](**kwargs)
+        else:
+            backend = BACKENDS['test']()
```
ZLLentz, line 88: These can also fail in a confusing way if the user puts an invalid backend string into their config, or if the test backend didn't load. I think we need to path the user's backend selection through a function with a clear error handling.

superscore/client.py
```diff
@@ -15,13 +19,124 @@ class Client:
...
+
+        # configure control layer and shims
+        if 'control_layer' in cfg_parser.sections():
+            shim_choices = [val for val, enabled
+                            in cfg_parser["control_layer"].items()
+                            if enabled]
+            control_layer = ControlLayer(shims=shim_choices)
+        else:
+            control_layer = ControlLayer()
```
ZLLentz, line 97: What's the default behavior here if no control layer is specified? Does the user need to be notified of this?

superscore/client.py
```diff
@@ -15,13 +19,124 @@ class Client:
...
+            cfg = cls.find_config()
+        if not os.path.exists(cfg):
+            raise RuntimeError(f"Superscore configuration file not found: {cfg}")
+
+        cfg_parser = configparser.ConfigParser()
+        cfg_file = cfg_parser.read(cfg)
+        logger.debug(f"Loading configuration file at ({cfg_file})")
+
+        # Gather Backend
+        if 'backend' in cfg_parser.sections():
```
ZLLentz, line 81: Note that, in general, user typos and extra sections in the config file will fail silently as implemented. Perhaps we need to be noisier for missing sections or alert on extra sections (e.g. if I make a `[bakned]` section or a similar typo I won't notice anything is wrong right away)

superscore/client.py
```diff
@@ -15,13 +19,124 @@ class Client:
...
+            Absolute path to the configuration file
+
+        Raises
+        ------
+        OSError
+            If no configuration file can be found by the described methodology
+        """
+        # Point to with an environment variable
+        if os.environ.get('SUPERSCORE_CFG', False):
+            happi_cfg = os.environ.get('SUPERSCORE_CFG')
```
ZLLentz, line 122: happi?

superscore/control_layers/core.py
```diff
@@ -19,10 +24,12 @@ class ControlLayer:
...
-    def __init__(self, *args, **kwargs):
-        self.shims = {
-            'ca': AiocaShim(),
-        }
+    def __init__(self, *args, shims: Optional[List[str]] = None, **kwargs):
+        if shims is None:
+            # load all available shims
+            self.shims = SHIMS
+        else:
+            self.shims = {key: shim for key, shim in SHIMS.items() if key in shims}
```
ZLLentz, line 32: Here's another place where a user including an invalid config will pass through silently, in this case a bad shim name.

ZLLentz (commented):
I think this is a good implementation but it isn't very helpful to a user making mistakes. We can do a lot more to alert someone that they've made a configuration error.

superscore/client.py
tangkong: I thought I caught them all darn

superscore/backends/__init__.py
tangkong: The short answer is yes, that is what I expected.  

This was set up with the intent that we might set up a `pip install superscore[mongo]` such that we could only install those extra dependencies if we wanted them.

superscore/client.py
tangkong: While that error path is a possibility, the setup in `_get_backends()` is entirely internal, and shouldn't depend on user input.  In particular I think we can guarantee the test Backend will exist.

superscore/client.py
tangkong: If no control layer is specified, we'll grab all the available shims.  I intended this to be the default behavior, and for the user to mainly configure the de-selection of shims.

I don't think we need to be loud about this, but I can add some debug statements

tangkong:
Good notes about being clearer about the default options.  I think I still want to not outright explode if some sections are missing (to give provide default behavior), but I've added more noisy logging output for sections that are less expected (missing backend specification)

ZLLentz (approved):

shilorigins (approved):

Issue #33, opened 2024-05-22
TST: unify test suite fixtures and interfaces


tangkong:
## Current Behavior
We have various test backend fixtures and unit tests, which should be made more easily usable in the future

## Expected Behavior
We've had tests come from two different styles/schools of thought, and we should unify them and make them universally usable

## Context / environment
From #25 and #27 

## Suggested Solution
Rewrite tests that expect specific uuids or field to work with the various backends?
Rewrite backend fixtures to return similar objects for use in other fixtures / tests (maybe have them all return `Root`)

PR #57, opened 2024-07-16
ENH: Implement snapshot saving

### LINKED ISSUE #32: Implement saving a Snapshot
## Description
Implement routine to take PVs detailed in a `Collection`, gather the PV data, and save that data into `Snapshots`

## Dependencies

- [ ] #31

shilorigins:
## Description

* Implement `Snapshot` saving (closes #32 )
  * include helper methods `Client._build_snapshot` and `Client._gather_pvs`
* Extend `Client._gather_data` to accept any type of `Entry`, and make iterative rather than recursive
## Motivation and Context

`Client.snap` itself is simple: it gathers all of the PVs in an entry, adds in the meta PVs, asynchronously fetches data for each PV, and then builds the snapshot. This method is named `snap` because `save` is currently used for the method that writes entries to the backend.

Since we want a function to gather all of the PVs from an entry, I decided to avoid code duplication by adding that functionality to `Client._gather_data`.  This extension needed some messy code to handle all of the parameter permutations, which I was able to simplify by changing the method to be iterative.  I also use `hasattr` rather than type checks to avoid length conditionals.  I included `Client._gather_pvs` as a convenience method.

Fetching the data is straightforward due to `ControlLayer._get_list`.

Building the snapshot uses a simple depth-first process.

### Alternative Designs
One major alternative was to have each `Entry` subclass implement method to gather its PVs. This conflicts with the current system because the entry classes don't have access to the backend to process UUIDs.
  
## How Has This Been Tested?

New and updated tests in `test_client.py`

This PR, docstrings

## Pre-merge checklist

### DIFF
--- a/superscore/client.py
+++ b/superscore/client.py
@@ -3,14 +3,17 @@ import configparser
 import logging
 import os
 from pathlib import Path
-from typing import Any, Generator, List, Optional, Union
+from typing import Any, Dict, Generator, List, Optional, Union
 from uuid import UUID
 
 from superscore.backends import get_backend
 from superscore.backends.core import _Backend
 from superscore.control_layers import ControlLayer
 from superscore.control_layers.status import TaskStatus
-from superscore.model import Entry, Setpoint, Snapshot
+from superscore.errors import CommunicationError
+from superscore.model import (Collection, Entry, Nestable, Parameter, Readback,
+                              Setpoint, Snapshot)
+from superscore.type_hints import AnyEpicsType
 from superscore.utils import build_abs_path
 
 logger = logging.getLogger(__name__)
@@ -163,6 +166,35 @@ class Client:
         """Compare two entries.  Should be of same type, and return a diff"""
         raise NotImplementedError
 
+    def snap(self, entry: Collection) -> Snapshot:
+        """
+        Asyncronously read data for all PVs under ``entry``, and store in a
+        Snapshot.  PVs that can't be read will have an exception as their value.
+
+        Parameters
+        ----------
+        entry : Collection
+            the Collection to save
+
+        Returns
+        -------
+        Snapshot
+            a Snapshot corresponding to the input Collection
+        """
+        logger.debug(f"Saving Snapshot for Collection {entry.uuid}")
+        pvs, _ = self._gather_data(entry)
+        pvs.extend(Collection.meta_pvs)
+        values = self.cl.get(pvs)
+        data = {}
+        for pv, value in zip(pvs, values):
+            if isinstance(value, CommunicationError):
+                logger.debug(f"Couldn't read value for {pv}, storing \"None\"")
+                data[pv] = None
+            else:
+                logger.debug(f"Storing {pv} = {value}")
+                data[pv] = value
+        return self._build_snapshot(entry, data)
+
     def apply(
         self,
         entry: Union[Setpoint, Snapshot],
@@ -194,7 +226,7 @@ class Client:
 
         # Gather pv-value list and apply at once
         status_list = []
-        pv_list, data_list = self._gather_data(entry)
+        pv_list, data_list = self._gather_data(entry, writable_only=True)
         if sequential:
             for pv, data in zip(pv_list, data_list):
                 logger.debug(f'Putting {pv} = {data}')
@@ -210,57 +242,109 @@ class Client:
 
     def _gather_data(
         self,
-        entry: Union[Setpoint, Snapshot, UUID],
-        pv_list: Optional[List[str]] = None,
-        data_list: Optional[List[Any]] = None
-    ) -> Optional[tuple[List[str], List[Any]]]:
+        entry: Union[Entry, UUID],
+        writable_only: bool = False,
+    ) -> tuple[List[str], Optional[List[Any]]]:
         """
-        Gather writable pv name - data pairs recursively.
-        If pv_list and data_list are provided, gathered data will be added to
-        these lists in-place. If both lists are omitted, this function will return
-        the two lists after gathering.
-
-        Queries the backend to fill any UUID values found.
+        Gather PV name - data pairs that are accessible from ``entry``.  Queries
+        the backend to fill any UUIDs found.
 
         Parameters
         ----------
-        entry : Union[Setpoint, Snapshot, UUID]
-            Entry to gather writable data from
-        pv_list : Optional[List[str]], optional
-            List of addresses to write data to, by default None
-        data_list : Optional[List[Any]], optional
-            List of data to write to addresses in ``pv_list``, by default None
+        entry : Union[Entry, UUID]
+            Entry to gather data from
+        writable_only : bool
+            If True, only include writable data e.g. omit Readbacks; by default False
 
         Returns
         -------
-        Optional[tuple[List[str], List[Any]]]
+        tuple[List[str], Optional[List[Any]]]
             the filled pv_list and data_list
         """
-        top_level = False
-        if (pv_list is None) and (data_list is None):
-            pv_list = []
-            data_list = []
-            top_level = True
-        elif (pv_list is None) or (data_list is None):
-            raise ValueError(
-                "Arguments pv_list and data_list must either both be provided "
-                "or both omitted."
+        pv_list = []
+        data_list = []
+
+        seen = set()
+        q = [entry]
+        while len(q) > 0:
+            entry = q.pop()
+            uuid = entry if isinstance(entry, UUID) else entry.uuid
+            if uuid in seen:
+                continue
+            elif isinstance(entry, UUID):
+                entry = self.backend.get_entry(entry)
+            seen.add(entry.uuid)
+
+            if isinstance(entry, Nestable):
+                q.extend(reversed(entry.children))  # preserve execution order
+            elif isinstance(entry, Readback) and writable_only:
+                pass
+            else:  # entry is Parameter, Setpoint, or Readback
+                pv_list.append(entry.pv_name)
+                if hasattr(entry, "data"):
+                    data_list.append(entry.data)
+                if getattr(entry, "readback", None) is not None:
+                    q.append(entry.readback)
+        return pv_list, data_list
+
+    def _build_snapshot(
+        self,
+        coll: Collection,
+        values: Dict[str, AnyEpicsType],
+    ) -> Snapshot:
+        """
+        Traverse a Collection, assembling a Snapshot using pre-fetched data
+        along the way
+
+        Parameters
+        ----------
+        coll : Collection
+            The collection being saved
+        values : Dict[str, AnyEpicsType]
+            A dictionary mapping PV names to pre-fetched values
+
+        Returns
+        -------
+        Snapshot
+            A Snapshot corresponding to the input Collection
+        """
+        snapshot = Snapshot(
+            title=coll.title,
+            tags=coll.tags.copy(),
+            origin_collection=coll
+        )
+
+        for child in coll.children:
+            if isinstance(child, UUID):
+                child = self.backend.get(child)
+            if isinstance(child, Parameter):
+                if child.readback is not None:
+                    readback = Readback(
+                        pv_name=child.readback.pv_name,
+                        description=child.readback.description,
+                        data=values[child.readback.pv_name]
+                    )
+                else:
+                    readback = None
+                setpoint = Setpoint(
+                    pv_name=child.pv_name,
+                    description=child.description,
+                    data=values[child.pv_name],
+                    readback=readback
+                )
+                snapshot.children.append(setpoint)
+            elif isinstance(child, Collection):
+                snapshot.append(self._build_snapshot(child, values))
+
+        snapshot.meta_pvs = []
+        for pv in Collection.meta_pvs:
+            readback = Readback(
+                pv_name=readback.pv_name,
+                data=values[readback.pv_name]
             )
+            snapshot.meta_pvs.append(readback)
 
-        if isinstance(entry, Snapshot):
-            for child in entry.children:
-                self._gather_data(child, pv_list, data_list)
-        elif isinstance(entry, UUID):
-            child_entry = self.backend.get_entry(entry)
-            self._gather_data(child_entry, pv_list, data_list)
-        elif isinstance(entry, Setpoint):
-            pv_list.append(entry.pv_name)
-            data_list.append(entry.data)
-
-        # Readbacks are not writable, and are not gathered
-
-        if top_level:
-            return pv_list, data_list
+        return snapshot
 
     def validate(self, entry: Entry):
         """
--- a/superscore/tests/conftest.py
+++ b/superscore/tests/conftest.py
@@ -681,6 +681,27 @@ def parameter_with_readback() -> Parameter:
 
 
 @pytest.fixture(scope='function')
+def setpoint_with_readback() -> Setpoint:
+    """
+    A simple setpoint-readback value pair
+    """
+    readback = Readback(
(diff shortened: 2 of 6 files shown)


superscore/client.py
```diff
@@ -160,6 +163,28 @@ def compare(self, entry_l: Entry, entry_r: Entry) -> Any:
...
+        Returns
+        -------
+        Snapshot
+            a Snapshot corresponding to the input Collection
+        """
+        logger.debug("Gathering PVs to read")
+        pvs = self._gather_pvs(entry)
+        pvs.extend(Collection.meta_pvs)
+        values = self.cl.get(pvs)
+        data = {pv: value if not isinstance(value, CommunicationError) else None for pv, value in zip(pvs, values)}
```
tangkong, line 185: I know we relaxed the line length limits on the pre-commit checks, but I'll argue once again for breaking lines into shorter limits (~80 characters is pretty standard).  Shorter lines help:
* people who need larger fonts
* people who like to look at code side-by-side
* limit the need to scroll side to side
* prevent line-wrap from doing weird things

The standard we try to follow is [pep8](https://peps.python.org/pep-0008/), the python guidance for style.  This helps other people who work on python code easily work in our codebase.  

I won't withhold an approval or anything petty, but I just wanted to argue for slightly shorter lines 😄

superscore/client.py
```diff
@@ -205,59 +230,106 @@ def apply(
...
-        Optional[tuple[List[str], List[Any]]]
+        tuple[List[str], Optional[List[Any]]]
             the filled pv_list and data_list
         """
-        top_level = False
-        if (pv_list is None) and (data_list is None):
-            pv_list = []
+        pv_list = []
+        data_list = None
+        if isinstance(entry, Snapshot) or isinstance(entry, Setpoint) or isinstance(entry, Readback):
```
tangkong, line 277:
```suggestion
        if isinstance(entry, (Snapshot, Setpoint, Readback)):
```

`isinstance` takes tuples as the second arg

superscore/tests/conftest.py
```diff
@@ -661,6 +661,46 @@ def sample_database() -> Root:
...
 
 
+@pytest.fixture(scope='function')
+def parameter_with_readback() -> Parameter:
+    """
+    A simple setpoint-readback parameter pair
+    """
+    readback = Parameter(
+        uuid="64772c61-c117-445b-b0c8-4c17fd1625d9",
+        pv_name="RBCK",
```
tangkong, line 671: Do the AD records use `RBCK`?  We typically see `RBV`.  (just a curiosity)

superscore/client.py
```diff
@@ -205,59 +230,106 @@ def apply(
...
+            if isinstance(child, Parameter):
+                if child.readback is not None:
+                    readback = Readback(
+                        pv_name=child.readback.pv_name,
+                        description=child.readback.description,
+                        data=values[child.readback.pv_name]
+                    )
+                else:
+                    readback = None
+                setpoint = Setpoint(pv_name=child.pv_name, description=child.description, data=values[child.pv_name], readback=readback)
```
tangkong, line 327: like you split the `Readback` definition into multiple lines above, why not this one? 😆

superscore/tests/test_client.py
```diff
@@ -53,6 +54,34 @@ def test_apply(put_mock, mock_client: Client, sample_database: Root):
     mock_client.apply(snap, sequential=True)
     assert put_mock.call_count == 3
 
+    put_mock.reset_mock()
+    mock_client.apply(setpoint_with_readback, sequential=True)
+    assert put_mock.call_count == 1
+
+
+@patch('superscore.control_layers.core.ControlLayer._get_one')
```
tangkong, line 62: Maybe it's worth `mock_client` also take the `dummy_cl`, and making the shim get/put methods `MagicMock` 🤔

superscore/tests/test_client.py
```diff
@@ -53,6 +54,34 @@ def test_apply(put_mock, mock_client: Client, sample_database: Root):
...
+    snapshot = mock_client.snap(coll)
+    assert get_mock.call_count == 5
+    assert all([snapshot.children[i].data == i for i in range(4)])  # children saved in order
+    setpoint = snapshot.children[-1]
+    assert isinstance(setpoint, Setpoint)
+    assert isinstance(setpoint.readback, Readback)
+    assert setpoint.readback.data == 4  # readback saved after setpoint
+
+
+@pytest.mark.parametrize('backends', [0], indirect=True)
```
tangkong, line 77: What are we using the backend here for?

superscore/client.py
```diff
@@ -205,59 +230,106 @@ def apply(
...
+            The collection being saved
+        values : Dict[str, AnyEpicsType]
+            A dictionary mapping PV names to pre-fetched values
+
+        Returns
+        -------
+        Snapshot
+            A Snapshot corresponding to the input Collection
+        """
+        snapshot = Snapshot(title="", origin_collection=coll, tags=coll.tags.copy())
```
tangkong, line 314: Did we want to guess at a title here?

superscore/tests/test_client.py
```diff
@@ -53,6 +54,34 @@ def test_apply(put_mock, mock_client: Client, sample_database: Root):
     mock_client.apply(snap, sequential=True)
     assert put_mock.call_count == 3
 
+    put_mock.reset_mock()
+    mock_client.apply(setpoint_with_readback, sequential=True)
+    assert put_mock.call_count == 1
+
+
+@patch('superscore.control_layers.core.ControlLayer._get_one')
+def test_snap(get_mock, mock_client, sample_database: Root, parameter_with_readback):
```
tangkong, line 63: Didn't want the other type hints for `mock_client` and `parameter_with_readback`?

superscore/client.py
shilorigins: I thought I tried this and got an error, but I think I must have tried a `Union`, which doesn't work with python 3.9.  Thanks for the catch

superscore/tests/test_client.py
shilorigins: I wanted to keep the signature shorter, and `mock_client` and `parameter_with_readback` contain their type hint in their name. I can make them explicit.

superscore/client.py
shilorigins: I was just looking at this. We've discussed having an autogen title, but not what it should contain.  I'm thinking of adding title, desc, and tags as parameters in `snap`, since the UI will prompt the user for these.

superscore/tests/test_client.py
tangkong: You can always break the signature into multiple lines 😉 

And also due to the indirect-ness of pytest, most linters/language servers won't pick up on the types unless you hint them explicitly.

tangkong: It's also worth disambiguating `parameter_with_readback` as just a `Parameter`, rather than the `tuple(Parameter, Readback)` that I thought it was

superscore/client.py
tangkong: We should remember that the client should be usable without the GUI, so we definitely shouldn't rely on user input (though I think it's worthwhile to give them the ability to).  I'm a fan of giving a default title to fall back on if nothing is provided.

shilorigins: Without a GUI, whoever's calling `snap` could still provide those args. I think it's sensible for the title to match the collection by default, with further context provided to the user by the UI or to a dev by the object type.

superscore/tests/test_client.py
shilorigins: To remember an earlier version of the test, apparently 🤦 . I initially tried serializing exceptions into the backend, but abandoned that for using `None` instead. The backend is no longer needed.

superscore/client.py
shilorigins: My wide monitor has made me blind to long lines apparently 😄

superscore/tests/conftest.py
shilorigins: I suddenly don't know, I think I made this up

superscore/client.py
shilorigins: I've split all lines longer than 95

superscore/tests/test_client.py
shilorigins: Hints added

superscore/tests/test_client.py
shilorigins: I'm happy to move the mock down a level

tangkong: I don't think it needs to happen in this PR.  That's more of a #33 followup than anything

superscore/client.py
```diff
@@ -208,59 +238,129 @@ def apply(
...
+                q.append(filled)
+            elif isinstance(entry, Nestable):
+                q.extend(reversed(entry.children))  # preserve execution order
+            elif isinstance(entry, Readback) and writable_only:
+                pass
+            else:  # entry is Parameter, Setpoint, or Readback
+                pv_list.append(entry.pv_name)
+                if hasattr(entry, "data"):
+                    data_list.append(entry.data)
+                if hasattr(entry, "readback") and entry.readback is not None:
```
tangkong, line 302:
```suggestion
                if getattr(entry, "readback", None) is not None:
```

superscore/client.py
```diff
@@ -208,59 +238,129 @@ def apply(
...
-        Optional[tuple[List[str], List[Any]]]
+        tuple[List[str], Optional[List[Any]]]
             the filled pv_list and data_list
         """
-        top_level = False
-        if (pv_list is None) and (data_list is None):
-            pv_list = []
+        pv_list = []
+        data_list = None
+        if isinstance(entry, (Snapshot, Setpoint, Readback)):
```
tangkong, line 285: We need to account for `entry` being a UUID, meaning we need to get the entry before initializing `data_list`

Is it even necessary to initialize `data_list` differently between read-only / read-write entries?  If we're not passed a (`Snapshot`, `Setpoint`, `Readback`), we just return an empty list.  (And then discard it as we do in `_gather_pvs`)

superscore/client.py
```diff
@@ -208,59 +238,129 @@ def apply(
...
+            tags=coll.tags.copy(),
+            origin_collection=coll
+        )
+
+        for child in coll.children:
+            if isinstance(child, UUID):
+                child = self.backend.get(child)
+            if isinstance(child, Parameter):
+                if child.readback is not None:
+                    readback = Readback(
```
tangkong, line 338: looking at this again, I've realized we don't record the status/severity here at all. 

Perhaps this requires a change to the payload we return from `ControlLayer.get`

superscore/client.py
```diff
@@ -208,59 +238,129 @@ def apply(
...
+        Returns
+        -------
+        List[str]
+            the list of PV names
         """
-        Gather writable pv name - data pairs recursively.
-        If pv_list and data_list are provided, gathered data will be added to
-        these lists in-place. If both lists are omitted, this function will return
-        the two lists after gathering.
+        pvs, _ = self._gather_data(entry, writable_only=False)
```
tangkong, line 259: Nitpick: This wrapper is maybe _too minimal_ to me, adding another layer of re-direction when someone in the future is trying to step through the methods in this code base.  I imagine it's not totally unreasonable to just take this line and use it above, dropping the empty data-list.

If it did some type-checking or error handling on top of `_gather_data`, I might be more on board

tangkong (commented):
Some more thoughts.  I think we need to think about how we handle `meta_pvs`, because at the very least, they are never initialized / configured in our current framework.

superscore/client.py
ZLLentz: fwiw I think `aioca` is actually returning you these `AugmentedValue` objects instead of just simple python data types. https://github.com/DiamondLightSource/aioca/blob/580177bf61d466c8166d5257988f93d08bd0d6ea/aioca/types.py#L96

tangkong: Can we make a follow up issue for this?

superscore/client.py
tangkong: I think this is the only change I really want to see, after which we can merge this and address the rest in followup efforts

superscore/client.py
```diff
@@ -163,6 +166,33 @@ def compare(self, entry_l: Entry, entry_r: Entry) -> Any:
...
+        logger.debug("Gathering PVs to read")
+        pvs = self._gather_pvs(entry)
+        pvs.extend(Collection.meta_pvs)
+        values = self.cl.get(pvs)
+        data = {}
+        for pv, value in zip(pvs, values):
+            if isinstance(value, CommunicationError):
+                data[pv] = None
+            else:
+                data[pv] = value
```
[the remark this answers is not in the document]
shilorigins: What do you think about only keeping the first instance of a PV in the lists returned by `_gather_data`? We can't use a set because we sometimes need to put values in order, but in those cases only the first occurrence of a PV needs to be written. Can you think of a reason `_gather_data` should include duplicate PVs in the first place?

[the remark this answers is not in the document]
tangkong: Oh no I don't think we want to hold duplicate entries for PVs.  

My point was more that if a rapidly chanigng PV shows up twice, there's no guarantee here that the most recent value will get used.  (If I recall) The way this is coded is that the deepest child will get written last.  

Since the list-get method is a collection of async coroutines, there's a chance that the deeper duplicate PV returns from its coroutine before a higher up one, making that value technically stale.

I think this is fine as is, and if we run into issues later we can revisit it

superscore/client.py
shilorigins: The case where it could be relevant is something like if a function calls `_gather_data` without checking `entry`'s type. If they expect a `Snapshot` and try to use `data_list`, but `entry` is actually a `Collection`, bugs are more likely if `data_list` is an empty list.  That's a fairly specific circumstance and should be caught earlier in the logic, but that was in the back of my mind.

superscore/client.py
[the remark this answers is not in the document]
tangkong: Ah sorry I misremembered.  It's the `.snap()` method here that places them into a dictionary here, rather than `_gather_data`.  

I don't think I have a strong opinion on this currently.  I think I favor letting later methods (`.snap()` or `.apply()`) handle these duplicates, for the sake of having `._gather_data()` do a simple and obvious job.  This also gives us flexibility later on

superscore/client.py
```diff
@@ -163,6 +166,33 @@ def compare(self, entry_l: Entry, entry_r: Entry) -> Any:
...
+        ----------
+        entry : Collection
+            the Collection to save
+
+        Returns
+        -------
+        Snapshot
+            a Snapshot corresponding to the input Collection
+        """
+        logger.debug("Gathering PVs to read")
```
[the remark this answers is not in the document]
shilorigins: I added the UUID, plus two more debug calls when reading each PV.

superscore/client.py
shilorigins: I've removed `_gather_pvs`.

superscore/client.py
shilorigins: I've changed `data_list` to always initialize to `[]`

superscore/client.py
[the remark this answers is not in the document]
shilorigins: I favor doing the work in `_gather_data`, since I can't think of a reason to keep duplicate PVs in a flat list without any grouping information, keeping dups requires duplicate work in calling functions, and I think `_gather_data`'s job stays intuitive. Since I've already implemented dedup-ing in `_gather_data`, and if your opinion isn't strong, I'll leave it like this for now.

superscore/client.py
shilorigins: Issue made (#71 )

tangkong (approved):
This looks good to me.  Let's get this in 👍

Issue #71, opened 2024-08-01
ENH: Record status and severity when reading PV values from EPICS


shilorigins:
## Current Behavior

We aren't tracking PV statuses or severities at all.
## Expected Behavior

`Setpoint`s and `Readback`s should represent the validity of their data through their `.status` and `.severity` attributes.
## Suggested Solution

If `aioca` is already returning `AugmentedValue` instances, we can retrieve this information from there. Either way, statuses and severities should probably be incorporated into the return value of `ControlLayer.get`.

tangkong:
Some digging revealed that while `aioca` claims it's returning an `AugmentedValue`, in reality the fields it populates depends on the format requested.  As ye ol' EPICS dictates, we have to specifically request controls fields from EPICS if we want them.  

For aioca, this involves `caget('PV:NAME', format=epicscorelibs.ca.dbr.FORMAT_TIME)`, which gives both timestamp and status/severity.

In the end we'll have to write a standardized interface type to smooth over differences between various backends.

PR #72, opened 2024-08-01
MNT: Common table views, Primarily LivePVTableModel, minor skeleton work on NestableTableModel

tangkong:
## Description
Build a couple table models that can hopefully be reused elsewhere.  Includes:
- `LivePVTableModel`: a model for displaying {Parameter, Readback, Setpoint} data alongside the current values
- `NestableTableModel`: a model for {Collection, Snapshot} entries, which shows title / description primarily
- `BaseTableModel` that holds their own entries

Also:
- refactors a bit of the existing table models
- Adds a `_PVPollThread` for grabbing data outside the GUI thread

Note: I renamed `superscore.widgets.tree` -> `superscore.widgets.views`, but the diff viewer doesn't pick that up.  I did not change the tree model at all, so the reviews can focus on the table models

To-do (in a later PR at this point):
- [ ] make columns hide-able (see #76)
- [x] Refine columns (also showing live status and severity?  This might wait on #71)
- [ ] Hierarchical headers to condense views (see #76)
- [x] Finesse poll callbacks and signals to minimally update the table

## Motivation and Context
Building screens, I fell down this rabbit hole and the diff warranted a separate PR

## How Has This Been Tested?
Interactively so far, a couple tests added

This PR

image

image

## Pre-merge checklist

### DIFF
--- a/superscore/tests/test_views.py
+++ b/superscore/tests/test_views.py
@@ -1,0 +1,55 @@
+from unittest.mock import MagicMock
+
+import pytest
+from pytestqt.qtbot import QtBot
+from qtpy import QtCore
+
+from superscore.client import Client
+from superscore.model import Parameter
+from superscore.widgets.views import LivePVTableModel
+
+
+@pytest.fixture(scope='function')
+def pv_poll_model(
+    mock_client: Client,
+    parameter_with_readback: Parameter,
+    qtbot: QtBot
+) -> LivePVTableModel:
+    model = LivePVTableModel(
+        client=mock_client,
+        entries=[parameter_with_readback],
+        poll_rate=1.0
+    )
+
+    # Make sure we never actually call EPICS
+    model.client.cl.get = MagicMock(return_value=1)
+    yield model
+
+    model.stop_polling()
+
+    qtbot.wait_until(lambda: not model._polling)
+
+
+def test_pvmodel_polling(pv_poll_model: LivePVTableModel, qtbot: QtBot):
+    thread = pv_poll_model._poll_thread
+    qtbot.wait_until(lambda: thread.running)
+
+    pv_poll_model.stop_polling()
+    qtbot.wait_until(lambda: thread.isFinished(), timeout=10000)
+    assert not thread.running
+
+
+def test_pvmodel_update(pv_poll_model: LivePVTableModel, qtbot: QtBot):
+    qtbot.wait_until(lambda: pv_poll_model._poll_thread.running)
+
+    assert pv_poll_model._data_cache
+
+    # make the mock cl return a new value
+    pv_poll_model.client.cl.get = MagicMock(return_value=3)
+
+    qtbot.wait_signal(pv_poll_model.dataChanged)
+
+    data_index = pv_poll_model.index_from_item(pv_poll_model.entries[0], 'Live Value')
+    qtbot.wait_until(
+        lambda: pv_poll_model.data(data_index, QtCore.Qt.DisplayRole) == '3'
+    )
--- a/superscore/tests/test_views.py
+++ b/superscore/tests/test_views.py
@@ -18,30 +18,27 @@ def pv_poll_model(
     model = LivePVTableModel(
         client=mock_client,
         entries=[parameter_with_readback],
-        poll_rate=1.0
+        poll_period=1.0
     )
 
     # Make sure we never actually call EPICS
     model.client.cl.get = MagicMock(return_value=1)
+    qtbot.wait_until(lambda: model._poll_thread.running)
     yield model
 
     model.stop_polling()
 
-    qtbot.wait_until(lambda: not model._polling)
+    qtbot.wait_until(lambda: not model._poll_thread.isRunning())
 
 
 def test_pvmodel_polling(pv_poll_model: LivePVTableModel, qtbot: QtBot):
     thread = pv_poll_model._poll_thread
-    qtbot.wait_until(lambda: thread.running)
-
     pv_poll_model.stop_polling()
     qtbot.wait_until(lambda: thread.isFinished(), timeout=10000)
     assert not thread.running
 
 
 def test_pvmodel_update(pv_poll_model: LivePVTableModel, qtbot: QtBot):
-    qtbot.wait_until(lambda: pv_poll_model._poll_thread.running)
-
     assert pv_poll_model._data_cache
 
     # make the mock cl return a new value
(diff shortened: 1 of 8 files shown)


tangkong:
I may work on #71 first... these tables want those data types, and I'll need to rebase this branch anyway

tangkong:
This got too long, so I'm cutting it off at the `LivePVTableModel` and leaving the `NestableTableModel` for another PR.  There's also still some improvements to be made for this table model but maybe in the spirit of an MVP we put that off

superscore/widgets/views.py
```diff
@@ -0,0 +1,954 @@
...
+    """
+    PV_Name = 0
+    Stored_Value = auto()
+    Live_Value = auto()
+    Timestamp = auto()
+    Stored_Status = auto()
+    Live_Status = auto()
+    Stored_Severity = auto()
+    Live_Severity = auto()
+    Open = auto()
```
ZLLentz, line 543: If these are going to be referenced in code as attributes and the names will be reprocessed anyway to sub in spaces, can we make these adhere to pep8 and do Title Case subbing later?

superscore/widgets/views.py
```diff
@@ -0,0 +1,954 @@
...
+        client: Client,
+        entries: Optional[List[PVEntry]] = None,
+        open_page_slot: Optional[Callable] = None,
+        poll_rate: float = 1.0,
+        **kwargs
+    ) -> None:
+        super().__init__(*args, entries=entries, **kwargs)
+
+        self.headers = [h.name.replace('_', ' ') for h in LivePVHeaderEnum]
+        self.HEADS = LivePVHeaderEnum
```
ZLLentz, line 569: What's up with this attribute? Why is it all caps and what purpose does it serve as an attribute vs referring to the original enum? Do we plan to be able to sub it out later?

superscore/widgets/views.py
```diff
@@ -0,0 +1,954 @@
...
+        self.client = client
+        self.open_page_slot = open_page_slot
+        self.poll_rate = poll_rate
+        self._data_cache = {e.pv_name: None for e in self.entries}  # noqa: F821
```
ZLLentz, line 573: Is `None` preferable here to some sort of disconnected/not connected yet `EpicsData` value?

superscore/widgets/views.py
```diff
@@ -0,0 +1,954 @@
...
+            data=self._data_cache,
+            poll_rate=self.poll_rate,
+            client=self.client
+        )
+
+        self._poll_thread.data_ready.connect(self._data_ready)
+        self._poll_thread.finished.connect(self._poll_thread_finished)
+
+        self._poll_thread.start()
+        self._polling = True
```
ZLLentz, line 594: The poll thread keeps track of its own running state, I think we should avoid duplicating that state here.

superscore/widgets/views.py
```diff
@@ -0,0 +1,954 @@
...
+        Returns
+        -------
+        QtCore.QModelIndex
+            The corresponding model index
+        """
+        row = self.entries.index(item)
+        if isinstance(column, int):
+            col = column
+        elif isinstance(column, str):
+            col = self.HEADS[column.replace(' ', '_')].value
```
ZLLentz, line 667: It might be valuable to have the enum itself take care of these conversions rather than re-implementing it in various places

superscore/widgets/views.py
```diff
@@ -0,0 +1,954 @@
...
+        Returns True if ``l_data`` is close to ``r_data``, False otherwise.
+        Intended for use with numeric values.
+        """
+        is_close = False
+        try:
+            is_close = np.isclose(l_data, r_data)
+        except TypeError:
+            pass
+
+        return is_close
```
ZLLentz, line 769: nitpick
```suggestion
        try:
            return np.isclose(l_data, r_data)
        except TypeError:
            return Flase
```

superscore/widgets/views.py
```diff
@@ -0,0 +1,954 @@
...
+    client : superscore.client.Client
+        The client to communicate to PVs through
+
+    data : dict[str, EpicsData]
+        Per-PV EpicsData, potentially generated previously.
+
+    poll_rate : float
+        The poll rate in seconds. A zero or negative poll rate will indicate
+        single-shot mode.  In "single shot" mode, the data is queried exactly
+        once and then the thread exits.
```
ZLLentz, line 804: nitpick: "rate" is typically a per-seconds speed measurement, maybe the docstring can clarify that that this is the time between successive polls

ZLLentz (commented):
I think this was all nitpicks and questions, there isn't anything bad here
Feel free to discard the ones you disagree with

superscore/tests/test_views.py
```diff
@@ -0,0 +1,55 @@
...
+    yield model
+
+    model.stop_polling()
+
+    qtbot.wait_until(lambda: not model._polling)
+
+
+def test_pvmodel_polling(pv_poll_model: LivePVTableModel, qtbot: QtBot):
+    thread = pv_poll_model._poll_thread
+    qtbot.wait_until(lambda: thread.running)
```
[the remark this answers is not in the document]
tangkong: There should be tests where this doesn't matter, but I haven't written them yet.  It probably does make sense to put this wait in the fixture

superscore/widgets/views.py
tangkong: Ya got me, I just didn't want to type `LivePVHeaderEnum`.  I'll rename some things to avoid this

superscore/widgets/views.py
tangkong: I thought I hit a non-TypeError at some point but I cannot reproduce it...

superscore/widgets/views.py
tangkong: This wasn't entirely intentional, `None` was just the easiest for me to initialize at the time.  It might helpful to be able to differentiate between the not-yet-attempted state and an attempted-but-failed state, but I'm not wedded to it.  

I suppose it might be better to keep the types consistent though

superscore/widgets/views.py
ZLLentz: A non-TypeError will raise in either variant here

tangkong: ... yes true (shame)

ZLLentz (approved):
LGTM

shilorigins (approved):

PR #73, opened 2024-08-05
ENH: Unify data handling under `EpicsData`

### LINKED ISSUE #71: ENH: Record status and severity when reading PV values from EPICS

tangkong:
## Description
Standardize our data handling to create and expect EpicsData instead of backend-specific data types.
Adjust tests and Client methods to comply with this change

## Motivation and Context
We weren't handling status and severity before, and we should do so in a unified way
closes #71 

Various libraries handle this EPICS metadata differently.  
* `aioca` defines an `AugmentedValue` protocol, which in python just means it's promising the returned value will have the same fields as `AugmentedValue`.  In reality the returned values are `epicscorelibs.ca.dbr.<ca_datatype>` types, which subclass python primitives and ONLY have fields dictated by the `format` kwarg
  * This means that even though the return value supposedly implements the `AugmentedValue` interface, it's not guaranteed to.  (tsk tsk)
* `pyepics` gets this information from the `cainfo` function, which calls `PV.get_ctrlvars` under the hood.  The metadata is then stored on the `PV` object (I believe)
* `p4p` seems to create a subclass of their own `NTBase` for each epics NT, and status/severity are stored as attributes on that object
  * These NT classes also subclass python primitives
  * confusingly, they define a `Type` class of their own to use throughout all this

In the end I think this is the minimal amount of work to get us to the place we want.  After we grab data from EPICS, we're dealing with normal data types stored in the `superscore.model` dataclasses, and won't be handling this custom shim layer anyway.  

## How Has This Been Tested?
Tests made to pass

This PR, and issue conversations

## Pre-merge checklist

### DIFF
--- a/superscore/client.py
+++ b/superscore/client.py
@@ -38,7 +38,7 @@ class Client:
         self.cl = control_layer
 
     @classmethod
-    def from_config(cls, cfg: Path = None):
+    def from_config(cls, cfg: Optional[Path] = None):
         """
         Create a client from the configuration file specification.
 
@@ -318,7 +318,9 @@ class Client:
                 child = self.backend.get(child)
             if isinstance(child, Parameter):
                 if child.readback is not None:
-                    edata = values[child.readback.pv_name] or EpicsData(data=None)
+                    edata = self._value_or_default(
+                        values.get(child.readback.pv_name, None)
+                    )
                     readback = Readback(
                         pv_name=child.readback.pv_name,
                         description=child.readback.description,
@@ -328,7 +330,7 @@ class Client:
                     )
                 else:
                     readback = None
-                edata = values[child.pv_name] or EpicsData(data=None)
+                edata = self._value_or_default(values.get(child.pv_name, None))
                 setpoint = Setpoint(
                     pv_name=child.pv_name,
                     description=child.description,
@@ -339,11 +341,11 @@ class Client:
                 )
                 snapshot.children.append(setpoint)
             elif isinstance(child, Collection):
-                snapshot.append(self._build_snapshot(child, values))
+                snapshot.children.append(self._build_snapshot(child, values))
 
         snapshot.meta_pvs = []
         for pv in Collection.meta_pvs:
-            edata = values[pv] or EpicsData(data=None)
+            edata = self._value_or_default(values.get(pv, None))
             readback = Readback(
                 pv_name=pv,
                 data=edata.data,
@@ -353,6 +355,12 @@ class Client:
             snapshot.meta_pvs.append(readback)
 
         return snapshot
+
+    def _value_or_default(self, value: Any) -> EpicsData:
+        """small helper for ensuring value is an EpicsData instance"""
+        if value is None or not isinstance(value, EpicsData):
+            return EpicsData(data=None)
+        return value
 
     def validate(self, entry: Entry):
         """
--- a/superscore/client.py
+++ b/superscore/client.py
@@ -8,12 +8,11 @@ from uuid import UUID
 
 from superscore.backends import get_backend
 from superscore.backends.core import _Backend
-from superscore.control_layers import ControlLayer
+from superscore.control_layers import ControlLayer, EpicsData
 from superscore.control_layers.status import TaskStatus
 from superscore.errors import CommunicationError
 from superscore.model import (Collection, Entry, Nestable, Parameter, Readback,
                               Setpoint, Snapshot)
-from superscore.type_hints import AnyEpicsType
 from superscore.utils import build_abs_path
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


ZLLentz:
I appreciate the thorough breakdown you wrote in the description here

superscore/control_layers/_base_shim.py
```diff
@@ -15,3 +23,41 @@ async def put(self, address: str, value: Any):
...
+@dataclass
+class EpicsData:
+    """Unified EPICS data type for holding data and relevant metadata"""
+    data: AnyEpicsType
+    status: Severity = Status.UDF
+    severity: Status = Severity.INVALID
+    timestamp: datetime = field(default_factory=utcnow)
+
+    @classmethod
+    def from_aioca(cls, value: AugmentedValue) -> EpicsData:
```
ZLLentz, line 37: Nitpick: this requires us to modify `_base_shim` when we want to change how a specific shim works, which feels like a break from general abstraction principles. Over time, `_base_shim` will need to have implementation details from every individual shim in it.

tangkong: I could put this in a different submodule, does that address the concern here?

I'm reading your comment as "the submodule `_base_shim` needs to change", not "the class `_BaseShim` needs to change"

ZLLentz: From an abstraction standpoint, I'd expect all the code that needs to know about aioca's API to be in the aioca shim, rather than putting some of it into the _base_shim module

tangkong: I've relocated this logic.  Good catch 👍

superscore/control_layers/_aioca.py
```diff
@@ -5,16 +5,19 @@
...
 from superscore.errors import CommunicationError
+from superscore.model import Severity, Status
 
 logger = logging.getLogger(__name__)
 
 
 class AiocaShim(_BaseShim):
     """async compatible EPICS channel access shim layer"""
-    async def get(self, address: str) -> Any:
+    async def get(self, address: str) -> EpicsData:
```
ZLLentz, line 20: as-implemented makes a ton of sense to me
would this type hint return signature also need to change on base shim for consistency?

ZLLentz (dismissed):
I think this is a good approach

superscore/control_layers/core.py
```diff
@@ -74,7 +75,7 @@ def shim_from_pv(self, address: str) -> _BaseShim:
         return shim
 
     @singledispatchmethod
-    def get(self, address: Union[str, list[str]]) -> Any:
+    def get(self, address: Union[str, list[str]]) -> Union[EpicsData, list[EpicsData]]:
```
shilorigins, line 78: Nitpick: we should keep our hints consistent between similar methods, and I favor using `Iterable` like we did for `_get_list`.

superscore/client.py
```diff
@@ -319,17 +318,23 @@ def _build_snapshot(
                 child = self.backend.get(child)
             if isinstance(child, Parameter):
                 if child.readback is not None:
+                    edata = values[child.readback.pv_name] or EpicsData(data=None)
```
shilorigins, line 321: If the PV name isn't in `values`, won't this will raise an error rather than returning `EpicsData(data=None)`?  Changing to `values.get(child.readback.pv_name, EpicsData(data=None))` should do what we want in one line (but the backup `EpicsData` instance will be created every time).

shilorigins (commented):
This mostly looks good, but I found one technical mistake and have a couple consistency suggestions.

superscore/control_layers/_aioca.py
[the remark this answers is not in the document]
tangkong: Ah yes correct all around.  Thank you for catching what my distracted brain did not

superscore/client.py
tangkong: Good catch.  To add onto this, our data gathering process inserts None for any values that cannot be found by EPICS.  This means we still need to check that the value is not None.  (None specifically, since empty strings or 0's are false-y in python_

superscore/control_layers/core.py
```diff
@@ -87,20 +88,20 @@ def get(self, address: Union[str, list[str]]) -> Any:
 
         Returns
         -------
-        Any
+        Union[EpicsData, list[EpicsData]]
```
ZLLentz, line 91: nitpick: should this return docstring be changed to Iterable too?

ZLLentz (approved):
I don't see any functional issues beyond what has already been reviewed/fixed, looks good

shilorigins (approved):

superscore/control_layers/core.py
tangkong: Man, this was really not my most detail oriented work haha.  I'll pick it up in the next PR, to stop bothering you all with this one

PR #77, opened 2024-09-03
MNT: read-only Parameters should be snapshotted as Readback, and propagate tolerances

tangkong:
## Description
If a `Parameter` in a `Collection` is marked as `read_only=True`, it should be captured as a `Readback` in the corresponding `Snapshot`.

Previously it would always be captured as a `Setpoint`, which is implicitly read-write, and did not have tolerance information

## Motivation and Context
Stumbled across this in another effort

## How Has This Been Tested?
I have not yet added a test

This PR

## Pre-merge checklist

### DIFF
--- a/superscore/client.py
+++ b/superscore/client.py
@@ -326,20 +326,34 @@ class Client:
                         description=child.readback.description,
                         data=edata.data,
                         status=edata.status,
-                        severity=edata.severity
+                        severity=edata.severity,
+                        rel_tolerance=child.readback.rel_tolerance,
+                        abs_tolerance=child.readback.abs_tolerance,
                     )
                 else:
                     readback = None
                 edata = self._value_or_default(values.get(child.pv_name, None))
-                setpoint = Setpoint(
-                    pv_name=child.pv_name,
-                    description=child.description,
-                    data=edata.data,
-                    status=edata.status,
-                    severity=edata.severity,
-                    readback=readback
-                )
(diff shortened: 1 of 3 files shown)


superscore/client.py
```diff
@@ -326,20 +326,34 @@ def _build_snapshot(
...
-                setpoint = Setpoint(
-                    pv_name=child.pv_name,
-                    description=child.description,
-                    data=edata.data,
-                    status=edata.status,
-                    severity=edata.severity,
-                    readback=readback
-                )
-                snapshot.children.append(setpoint)
+                if child.read_only:
```
ZLLentz, line 336: Is it ever possible for a normally constructed child `Parameter` to have a `readback` but also be `read_only`? If so, this ends up removing all references to the original `readback` but I'm not sure if this ever happens at all.

ZLLentz (approved):
One question from me but the result and implementation are sensible

shilorigins (approved):

superscore/client.py
shilorigins: I don't think so.  An exception could be meta PVs, but we currently keep track of them specifically and so don't set the flag.  As it stands, I don't think there's a use case where a read-only setpoint would have a readback.  In fact, the `read_only` attribute isn't currently used, so its only purpose is to distinguish between `Setpoint`s and `Readback`s.

Issue #74, opened 2024-08-09
MNT/BUG: single `_AiocaShim.get('pv')` should return a CommunicationError in the same way as list version


tangkong:
## Current Behavior
Currently the single-pv version raises the communication error, while the list version returns a list of such errors.  

## Expected Behavior
The two should act similarly

## Context / environment
poking around, this is something I saw in the past but never ironed out

## Steps to Reproduce (for bugs)
1. `from superscore.control_layers import ControlLayer; cl = ControlLayer()`
2. `res_single = cl.get('nonsense:non:working:pv')`  (should raise)
3. `res_list = cl.get(['nonsense', 'nonsense2'])`  (should return list)
4.  Compare `res_single`, `res_list`

## Suggested Solution
Catch the exception in the single version and return it>

PR #78, opened 2024-09-05
MNT: return error rather than raising in ControlLayer._get_single

shilorigins:
## Description

Adjust `_get_single` to intercept and return a `CommunicationError` rather than propagating it.

`_get_one` still raises the error so that alternative `get` methods will still propagate the error unless they make the decision to return it.

## Motivation and Context

Context in #74 
## How Has This Been Tested?

New test: `test_cl.py::test_get_communication_error`

This PR
## Pre-merge checklist

### DIFF
--- a/superscore/control_layers/core.py
+++ b/superscore/control_layers/core.py
@@ -10,6 +10,7 @@ from typing import Any, Callable, Dict, List, Optional, Union
 
 from superscore.control_layers._base_shim import EpicsData
 from superscore.control_layers.status import TaskStatus
+from superscore.errors import CommunicationError
 
 from ._aioca import AiocaShim
 from ._base_shim import _BaseShim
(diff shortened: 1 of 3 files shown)


superscore/control_layers/core.py
```diff
@@ -98,7 +99,11 @@ def get(self, address: Union[str, Iterable[str]]) -> Union[EpicsData, Iterable[E
     @get.register
     def _get_single(self, address: str) -> EpicsData:
```
ZLLentz, line 100: My only suggestion is to update the return value in the signature here
```suggestion
    def _get_single(self, address: str) -> EpicsData | CommunicationError:
```

superscore/control_layers/core.py
```diff
@@ -98,7 +99,11 @@ def get(self, address: Union[str, Iterable[str]]) -> Union[EpicsData, Iterable[E
...
         """Synchronously get a single ``address``"""
-        return asyncio.run(self._get_one(address))
+        try:
+            res = asyncio.run(self._get_one(address))
+        except CommunicationError as e:
+            res = e
+        return res
```
ZLLentz, line 106: One more silly nitpick, no need to do any variable assignments here
```suggestion
        try:
            return asyncio.run(self._get_one(address))
        except CommunicationError as e:
            return e
```

tangkong (commented):
I'm on board with this, nothing to add past what Zach has mentioned

superscore/control_layers/core.py
shilorigins: Easy to forget, good catch

tangkong (approved):

ZLLentz (approved):

PR #79, opened 2024-09-09
ENH: Augment search methods

### LINKED ISSUE #61: ENH: Augment search methods / terms
## Current Behavior
Currently search methods on the `_Backend` classes only look for exact matches between key-value pairs on `Entry` and its subclassses.  

## Expected Behavior
We should be able to handle slightly more complex search queries, including but not limited to:
- substring / regex matching
- greater-than / less-than conditions (for dates, etc)
- matching if the value is in a list of provided values
  - for example, in matching tags?
- matching values in meta_pv list?
- more I'm sure 

## Context / environment
Some of this was added in #56, but this deserves its own effort

N/A

## Suggested Solution
This makes sense to implement in the `_Backend` classes, since different backends might have different shortcuts for these queries (e.g. mongo has $lt, but filestore backend will have to do the comparison in python)

One thought is to create separate search methods, for specific types of searches (`search_regex`, `search_range`).  This requires combining the different methods for the sake of presenting a single interface for users.  This would be cleaner from a modularity point of view, and also make it clear what needs to be implemented in `_Backend` subclasses

shilorigins:
## Description

* Change search term format from keyword arg to namedtuple of `(<attribute>, <operator>, <value>)`
* Define `SearchTerm` namedtuple
* Modify search page, views, and tests to use new format

Closes #61 
### SearchTerm Format
A three-term format provides more flexibility than a keyword arg. For example, we'd run into trouble using keyword args if we wanted to support both exact match and fuzzy match on entry titles, but with the tuple format we can apply different operators to the same attribute / value pairs.

Using a namedtuple for the format provides the option of setting and accessing args by name without taking more memory than bare tuples.  

### Operators
I included the operators:
* `eq` (equals)
* `lt` (less than or equal to)
* `gt` (greater than or equal to)
* `in` (item in container)
* `like` (partial match)
* `isclose` (number within absolute and relative tolerances)

`lt` and `gt` are both inclusive because I felt that that was more intuitive, and that we didn't need to also support exclusive versions. `like` currently supports regex string and UUID matches.

`isclose` takes an arg with format `(<value>, <rel_tolerance>, <abs_tolerance>)`.  The client converts this `SearchTerm` to `lt` and `gt`; the intention is that tolerances can be taken from the user or from the `Entry` directly, which the client turns into upper and lower bounds.

`entry_type` requires special handling because type comparisons need subclass information.

### Tags
`FilestoreBackend` implicitly supports tags using the `gt` operator, but a dedicated operator may be required when other backends are added. To find entries whose tag set contains all tags in a filter, use the conditional `('tags', 'gt', {<tags>...})`, which calls `set.__ge__`.  Supporting more complex tag filters, like inversion or disjunction, will also require a tag operator or using `Flag` functionality.

## How Has This Been Tested?

Existing tests in `test_page`, `test_window`, and `test_backend`.
New tests for `lt`, `gt`, fuzzy match, and tag filters.

This PR

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/79-augment_search_methods.rst
+++ b/docs/source/upcoming_release_notes/79-augment_search_methods.rst
@@ -1,0 +1,24 @@
+79 augment search methods
+#################
+
+API Breaks
+----------
+- Client.search takes SearchTerms as *args rather than key-value pairs as **kwargs
+
+Features
+--------
+- regex search on Entry text fields
+- filter Entrys by tag
+- filter Entrys by attribute value
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- shilorigins
--- a/superscore/backends/core.py
+++ b/superscore/backends/core.py
@@ -1,13 +1,21 @@
 """
 Base superscore data storage backend interface
 """
-from collections import namedtuple
-from typing import Generator
+from collections.abc import Container, Generator
+from typing import NamedTuple, Union
 from uuid import UUID
 
 from superscore.model import Entry, Root
+from superscore.type_hints import AnyEpicsType
 
-SearchTerm = namedtuple('SearchTerm', ('attr', 'operator', 'value'))
+SearchTermValue = Union[AnyEpicsType, Container[AnyEpicsType], tuple[AnyEpicsType, ...]]
+SearchTermType = tuple[str, str, SearchTermValue]
+
+
+class SearchTerm(NamedTuple):
+    attr: str
+    operator: str
+    value: SearchTermValue
 
 
 class _Backend:
@@ -44,7 +52,7 @@ class _Backend:
         """
         raise NotImplementedError
 
-    def search(self, *search_terms) -> Generator[Entry, None, None]:
+    def search(self, *search_terms: SearchTermType) -> Generator[Entry, None, None]:
         """
         Yield Entry objects matching all ``search_terms``. Each SearchTerm has the format
         (<attr>, <operator>, <value>).  Some operators take tuples as values.
--- a/superscore/backends/core.py
+++ b/superscore/backends/core.py
@@ -1,16 +1,20 @@
 """
 Base superscore data storage backend interface
 """
+from collections import namedtuple
 from typing import Generator
 from uuid import UUID
 
 from superscore.model import Entry, Root
+
+SearchTerm = namedtuple('SearchTerm', ('attr', 'operator', 'value'))
 
 
 class _Backend:
     """
     Base class for data storage backend.
     """
+
     def get_entry(self, meta_id: UUID) -> Entry:
         """
         Get entry with ``meta_id``
@@ -40,8 +44,18 @@ class _Backend:
         """
         raise NotImplementedError
 
-    def search(self, **search_kwargs) -> Generator[Entry, None, None]:
-        """Yield a Entry objects corresponding matching ``search_kwargs``"""
+    def search(self, *search_terms) -> Generator[Entry, None, None]:
+        """
+        Yield Entry objects matching all ``search_terms``. Each SearchTerm has the format
+        (<attr>, <operator>, <value>).  Some operators take tuples as values.
+
+        The supported operators are:
+        - eq (equals)
+        - lt (less than or equal to)
+        - gt (greater than or equal to)
+        - in
+        - like (fuzzy match, depends on type of value)
+        """
         raise NotImplementedError
 
     @property
--- a/superscore/backends/filestore.py
+++ b/superscore/backends/filestore.py
@@ -293,33 +293,22 @@ class FilestoreBackend(_Backend):
         """
         with self._load_and_store_context() as db:
             for entry in db.values():
-                match_found = True
+                conditions = []
                 for key, value in search_kwargs.items():
-                    # specific type handling, assuming is tuple
-                    if key == "entry_type":
-                        if not isinstance(entry, search_kwargs["entry_type"]):
-                            match_found = False
-
+                    if key == "entry_type":  # specific type handling, assume value is tuple
+                        conditions.append(isinstance(entry, value))
                     elif key == "start_time":
-                        if value > entry.creation_time:
-                            match_found = False
+                        conditions.append(entry.creation_time > value)
                     elif key == "end_time":
-                        if entry.creation_time > value:
-                            match_found = False
-
+                        conditions.append(entry.creation_time < value)
                     # TODO: search for child pvs?
-
-                    # plain key-value match
-                    else:
+                    else:  # plain key-value match
                         entry_value = getattr(entry, key, None)
                         if isinstance(value, tuple):
-                            matched = entry_value in value
+                            conditions.append(entry_value in value)
                         else:
-                            matched = entry_value == value
-
-                        match_found = match_found and matched
-
-                if match_found:
+                            conditions.append(entry_value == value)
+                if all(conditions):
                     yield entry
 
     @contextlib.contextmanager
--- a/superscore/backends/filestore.py
+++ b/superscore/backends/filestore.py
@@ -337,9 +337,9 @@ class FilestoreBackend(_Backend):
         elif op == "in":
             return data in target
         elif op == "like":
-            if isinstance(data, str):
-                return re.search(target, data)
-        return NotImplemented
+            return re.search(target, data)
+        else:
+            raise ValueError(f"SearchTerm does not support operator \"{op}\"")
 
     @contextlib.contextmanager
     def _load_and_store_context(self) -> Generator[Dict[UUID, Any], None, None]:
--- a/superscore/backends/filestore.py
+++ b/superscore/backends/filestore.py
@@ -338,6 +338,8 @@ class FilestoreBackend(_Backend):
         elif op == "in":
             return data in target
         elif op == "like":
+            if isinstance(data, UUID):
+                data = str(data)
             return re.search(target, data)
         else:
             raise ValueError(f"SearchTerm does not support operator \"{op}\"")
--- a/superscore/backends/filestore.py
+++ b/superscore/backends/filestore.py
@@ -6,6 +6,7 @@ import contextlib
 import json
 import logging
 import os
+import re
 import shutil
 from dataclasses import fields, replace
 from typing import Any, Dict, Generator, Optional, Union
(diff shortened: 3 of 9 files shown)


docs/source/upcoming_release_notes/79-augment_search_methods.rst
```diff
@@ -0,0 +1,24 @@
+79 augment search methods
+#################
+
+API Breaks
+----------
+- N/A
```
tangkong, line 6: This is technically an api break, in `_Backend.search`

superscore/client.py
```diff
@@ -147,9 +147,24 @@ def find_config() -> Path:
         # If found nothing
         raise OSError("No superscore configuration file found. Check SUPERSCORE_CFG.")
 
-    def search(self, **post) -> Generator[Entry, None, None]:
-        """Search by key-value pair.  Can search by any field, including id"""
-        return self.backend.search(**post)
+    def search(self, *post) -> Generator[Entry, None, None]:
```
tangkong, line 150: We should think about how to make this user friendly, or if we need to.  I envisioned other applications making calls to `Client.search`, and worry the extra cognitive load of constructing proper `SearchTerm`'s might discourage people.  

Some thoughts / proposals:
* We retain back-compat by **also** supporting key-value matching at the client level, but translate any we get into `SearchTerm`'s for `_Backend` methods.  
* We add some type-checking or factory methods that make creating a `SearchTerm` less error prone.  I really like the simple tuple interface, but maybe we have a `NamedTuple` subclass that raises descriptive errors.

superscore/backends/core.py
```diff
@@ -40,8 +44,18 @@ def update_entry(self, entry: Entry) -> None:
...
 
-    def search(self, **search_kwargs) -> Generator[Entry, None, None]:
-        """Yield a Entry objects corresponding matching ``search_kwargs``"""
+    def search(self, *search_terms) -> Generator[Entry, None, None]:
+        """
+        Yield Entry objects matching all ``search_terms``. Each SearchTerm has the format
+        (<attr>, <operator>, <value>).  Some operators take tuples as values.
+
+        The supported operators are:
+        - eq (equals)
```
tangkong, line 53: I thought that this might be nice as an `Enum`, but it'd have to be an enum with a short name to avoid pain while typing out a search query...

superscore/tests/test_backend.py
```diff
@@ -73,42 +74,99 @@ def test_delete_entry(backends: _Backend):
...
     # Search by field name
-    results = backends.search(data=2)
+    results = backends.search(
+        SearchTerm('data', 'eq', 2)
+    )
     assert len(list(results)) == 3
     # Search by field name
     results = backends.search(
-        uuid=UUID('ecb42cdb-b703-4562-86e1-45bd67a2ab1a'), data=2
+        SearchTerm('uuid', 'eq', UUID('ecb42cdb-b703-4562-86e1-45bd67a2ab1a')),
```
tangkong, line 92: I think this gives us the opportunity to permit searching sub-strings if comparing to a uuid.  Or maybe we should just start using the 'in' operator for partial UUID matches

tangkong (commented):
I like this a lot.  I think it gives us a lot of opportunity to make some robust search functionality.  

I do have some concerns about user friendliness at the Client-level, but I think everything behind that (backend, GUI) is fine as written.

superscore/backends/core.py
shilorigins: I had the same thought, but I didn't come up with a name I liked.

superscore/tests/test_backend.py
shilorigins: Is there a situation where a user would only have part of a UUID they want to match? UI users won't interact with UUIDs directly, and client users should only have access to whole UUIDs, so I'm not sure the benefit of partial matches.

After some quick interactive testing, UUIDs are not iterable and don't work with `re.search`, so they'd take special handling to work with the current `in` and `like` operators.

superscore/client.py
shilorigins: I didn't know about `typing.NamedTuple`, I'll at least refactor `SearchTerm` to use that, since it's equivalent but enables type hints.

I tried to thread a needle between simplicity, intuition, and completeness, but I definitely might have missed.  What mistakes do you envision when users create `SearchTerm`s?  They can be instantiated using positional or keyword args, and factory methods would still require one or the other. Type checking seems less maintainable, but I'm not sure how you're envisioning it working.

I'm open to also supporting key-value pairs in `Client.search`, but only `eq` would be supported. If a significant enough proportion of filters use only that operator then it would be worth supporting, but I think it could also muddy the interface. I guess it's a question of how annoying users find `SearchTerm` to be.

superscore/tests/test_backend.py
tangkong: I was more thinking of how git lets me specifiy a commit with only the first few characters of the commit hash.  I wouldn't put this at the highest priority, more of a thought that it might be nice for us to not have to make tests with the whole uuid string, rather something like:
```
SearchTerm('uuid', 'eq', 'ecb42cdb)  == SearchTerm('uuid', 'eq', UUID('ecb42cdb-b703-4562-86e1-45bd67a2ab1a')
```

superscore/client.py
tangkong: I thought of factory methods more as an entrypoint for us to add helpful error messages or slight corrective argument mangling.

shilorigins: Thanks, that helps.

shilorigins: The logic is currently set so that filters evaluate to `False` if there are mistakes in the attributes or operators, but this silent handling doesn't suit client use, so I'll add checks and exceptions.

superscore/backends/core.py
```diff
@@ -40,8 +44,18 @@ def update_entry(self, entry: Entry) -> None:
         """
         raise NotImplementedError
 
-    def search(self, **search_kwargs) -> Generator[Entry, None, None]:
-        """Yield a Entry objects corresponding matching ``search_kwargs``"""
+    def search(self, *search_terms) -> Generator[Entry, None, None]:
```
ZLLentz, line 47: Small thing: many of the signatures in this PR avoid type hinting the search terms.
I think it wouldn't be too hard to hint these via alias:
````
SearchValueType = AnyEpicsType | tuple[AnyEpicsType, ...]
SearchTermsType = tuple[str, str, SearchValueType]
````

File: superscore/client.py [hunk 6]
```diff
@@ -147,9 +147,24 @@ def find_config() -> Path:
...
+        """
+        Search backend for entries matching all SearchTerms in ``post``.  Can search by any
+        field, plus some special keywords. Backends support operators listed in _Backend.search.
+        Some operators are supported in the UI / client and must be converted before being
+        passed to the backend.
+        """
+        new_search_terms = []
+        for search_term in post:
+            if search_term.operator == 'like_with_tols':
```
ZLLentz, line 159: Since `like` means fuzzy search, I think `like_with_tols` might be a confusing name for the user, creating mix-ups between fuzzy text search and numeric range searching.

Some suggestions for renames:

- within_tols
- close_to
- isclose (from numpy)

I don't mind so much if one of these or some other name is used but I think the name should be distinct from `like`.

ZLLentz (commented):
I really like the flexibility this gives us in the search

superscore/client.py
shilorigins: The original idea was for `like` to work differently for different arg types.  It would always mean "similar to", but for strings that could mean fuzzy search whereas for numbers it could mean `isclose`.  That's an even more direct name collision, but it would group similar functionality (at least in my mind) under one name.

I decided to go with more explicit operators, so I'll rename `like_with_tols` rather than folding it into `like`, but I wanted to make a note of the idea.

superscore/client.py
shilorigins: You can now pass bare tuples as args to `Client.search`, so the following is valid:
```
result = client.search(
    ("creation_time", "gt", "2023-09-01"),
    ("creation_time", "lt", "2024-09-01"),
    ("pv_name", "in", "2023-09-01"),
    ("description", "like", "my_desc"),
)
```
Hopefully that helps verbosity.

The filestore backend now raises a `ValueError` if `op` is not one of the supported strings.  Instead of raising its own `TypeError`, I'm relying on the errors from the underlying functions if `op` doesn't support the received arg types.  This is less explicitly clear, but saves type checking that's repeated in the underlying functions, and the stack trace should be enough for client users to track errors.

Typos in `attr` still produce empty results.  It's nice for `Entry`s to automatically be filtered out if they don't have `attr`, but maybe I could check if `attr` is valid for any entry type before proceeding, so that typos could be caught.

I'm also still noodling on factory methods.

superscore/backends/core.py
shilorigins: It's a bit funky.

The client can take in tuples and converts them to `SearchTerm`s, but the backend search methods can currently also take tuples if called directly, so it makes sense to hint `SearchTermType`.

Similarly, `SearchTermValue` only needs to include `tuple[AnyEpicsType, ...]` for `Client.search`, because none of the current operators take a tuple as a target value, but future operators might require tuples.  Additionally, `SearchTermValue` should include `collections.abc.Container` because of `in`, but that's kind of specific to `FilestoreBackend` using Python's `__contains__` magic method.

I went with broad definitions, but I expect the type hints will change as we add concrete backend classes and have a better idea of what methods and object types are actually needed.

superscore/client.py
shilorigins: I think it's best to leave `attr` validation for a future PR. I've come up with a couple approaches, but I'm not happy with two of them and the third has a significant footprint. I've created #83 to track this if we want to implement it in the future.

superscore/backends/core.py
```diff
@@ -1,14 +1,14 @@
...
-from collections.abc import Container
-from typing import Generator, NamedTuple
+from collections.abc import Container, Generator
+from typing import NamedTuple, Union
 from uuid import UUID
 
 from superscore.model import Entry, Root
 from superscore.type_hints import AnyEpicsType
 
-SearchTermValue = AnyEpicsType | Container[AnyEpicsType] | tuple[AnyEpicsType, ...]
```
tangkong: FWIW if you really like the `|` union, you can invoke `from __future__ import annotations` at the top of the file.

superscore/tests/test_backend.py
shilorigins: I've enabled searching for substrings of UUIDs in `like`. Since `eq` is meant for exact matches and `like` is meant for close-enough matches, it felt more natural there.

superscore/backends/core.py
shilorigins: Thanks for the heads up.  I don't mind either way, I'm happy to stick to <=3.9 syntax.

superscore/tests/test_backend.py
```diff
@@ -73,42 +74,118 @@ def test_delete_entry(backends: _Backend):
...
-    results = backends.search(entry_type=(Snapshot, Collection))
+    results = backends.search(
+        SearchTerm('entry_type', 'in', (Snapshot, Collection))
+    )
     assert len(list(results)) == 2
 
+    results = backends.search(
+        SearchTerm('data', 'lt', 3)
+    )
+    assert len(list(results)) == 3
```
ZLLentz, line 110: Something I hadn't considered in the previous PRs: would it add value to check that the results match the searches in the test suite? Or is this not worth the effort? I'm not sure but I thought I'd mention it.
```
for res in results:
    assert res.data <= 3
```

ZLLentz (approved):
This looks clean to me. I had one side thought but do not take it as a call to action.

docs/source/upcoming_release_notes/79-augment_search_methods.rst
```diff
@@ -0,0 +1,24 @@
+79 augment search methods
+#################
+
+API Breaks
+----------
+- Client.search takes SearchTerms as *args rather than key-value pairs as **kwargs
```
tangkong, line 6: The github diff rendered part of this as italics... strange

superscore/backends/core.py
```diff
@@ -1,16 +1,28 @@
...
 """
-from typing import Generator
+from collections.abc import Container, Generator
+from typing import NamedTuple, Union
 from uuid import UUID
 
 from superscore.model import Entry, Root
+from superscore.type_hints import AnyEpicsType
+
+SearchTermValue = Union[AnyEpicsType, Container[AnyEpicsType], tuple[AnyEpicsType, ...]]
```
tangkong, line 11: I've never actually used `Container` as a type hint, kinda neat 👍 .
  Is a tuple not covered by `Container?`  It counts as an instance to me (`isinstance((1,2,3), Container) == True`)

superscore/backends/filestore.py
```diff
@@ -284,43 +286,63 @@ def delete_entry(self, entry: Entry) -> None:
...
-                        entry_value = getattr(entry, key, None)
-                        if isinstance(value, tuple):
-                            matched = entry_value in value
-                        else:
-                            matched = entry_value == value
+                        try:
+                            # check entry attribute by name
+                            value = getattr(entry, attr)
+                            conditions.append(self.compare(op, value, target))
+                        except AttributeError:
```
tangkong, line 307: Just thinking about the failure paths here:
* operator not supported -> `compare()` throws a descriptive exception
* attribute not present -> silent omission of the criteria

I think I'm ok with the behavior being the omission of the search term from the query, but we should at least add a logger.debug message (maybe the louder logger.info is appropriate)

superscore/backends/filestore.py
```diff
@@ -284,43 +286,63 @@ def delete_entry(self, entry: Entry) -> None:
...
-                        if entry.creation_time > value:
-                            match_found = False
-
+                conditions = []
+                for attr, op, target in search_terms:
                     # TODO: search for child pvs?
-
-                    # plain key-value match
+                    if attr == "entry_type":
+                        conditions.append(isinstance(entry, target))
```
tangkong, line 301: So in the 'entry_type' case, it actually doesn't matter what operator we choose?  (This is fine, just clarifying)

tangkong (dismissed):
I'm onboard, I like this implementation a lot.

A few nitpicky comments from me, none of which should hold up merging of this PR

docs/source/upcoming_release_notes/79-augment_search_methods.rst
shilorigins: It behaves weird with my vim highlighting too.  Maybe the double `**` is a parsing edge case?

tangkong: image

It renders funny in sphinx too, the `*` ends up being a hyperlink to an id anchor?  This is something we can fix up when we actually publish release notes though

superscore/backends/core.py
shilorigins: `Container` does cover `tuple`, but I preferred making the three cases explicit because they cover different cases.  `AnyEpicsType` covers operators that use singular data, `tuple` covers operators that use multiple data, and `Container` only covers `in` (although this is specific to `FilestoreBackend`'s implementation and may be reconsidered when we add a db backend).

superscore/backends/filestore.py
shilorigins: Correct.  We can distinguish operators in the future if we want to, for example if we want `eq` to require exact matches and `like` to cover multiple options or subclasses, but `isinstance` covers both cases automatically and seems reasonable.

superscore/backends/filestore.py
shilorigins: Case two is little different than you describe. When `getattr` raises an `AttributeError`, that means the current entry doesn't have `attr` and so the entry fails the search term and is omitted from the search results. The search term with `attr` is still checked for later entries though.

This is nice, because otherwise a search term like `('data', 'eq', 5)` would log a message for every `Parameter` and `Collection` in the backend.

docs/source/upcoming_release_notes/79-augment_search_methods.rst
ZLLentz: Later we can add backticks around these to force them to be literals

superscore/backends/filestore.py
shilorigins: My comment above might be too specific to the highlighted code. There's also the sub-case where `attr` isn't present in any entry, such as for a typo.  In that case, the search results will be empty and no warning is given.  #83 has ideas for when to warn the user.

tangkong: Sounds good to me, and suitable for coverage under #83

shilorigins:
I had to edit the collection builder page merged in earlier today, so I appreciate your re-approval 🙏

tangkong (approved):

PR #80, opened 2024-09-11
ENH: add FilterComboBox, CollectionBuilderPage

### LINKED ISSUE #49: GUI: Build Collection builder page
Some view where users can add PVs (Parameters) and adjust their metadata

Must have methods for adding children collections

tangkong:
## Description
- adds CollectionBuilderPage, with fields/forms for adding Parameters, sub-Collections to a collection
  - Utilizes shared `LivePVTableModel` and `NestableModel`
  - Adds a `FilterableComboBox`

To-Do:
- [x] Improve model update lag time, avoiding gc errors when stopping polling
- [x] Have tree-view update when changes are made
- [x] enable delete
- [x] Write tests
- [x] make accessible from central ui

## Motivation and Context
closes #49 

## How Has This Been Tested?
Interactively, tests to be added

This PR.

image

## Pre-merge checklist

### DIFF
--- a/superscore/widgets/page/collection_builder.py
+++ b/superscore/widgets/page/collection_builder.py
@@ -57,6 +57,7 @@ class CollectionBuilderPage(Display, DataWidget):
         self.open_page_slot = open_page_slot
         self.pv_model = None
         self.coll_model = None
+        self.tree_model = None
         self._coll_options: list[Collection] = []
         self._title = self.data.title
         # TODO: fill uuids here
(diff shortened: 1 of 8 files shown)


superscore/widgets/page/collection_builder.py
```diff
@@ -0,0 +1,236 @@
...
+        self.tree_view.setModel(self.tree_model)
+        # initialize tables
+        self.sub_colls = [child for child in self.data.children
+                          if isinstance(child, Collection)]
+        self.sub_pvs = [child for child in self.data.children
+                        if not isinstance(child, Collection)]
+        if self.pv_model is not None:
+            logger.debug('stopping polling')
+            self.pv_model.stop_polling()
+            self.pv_model._poll_thread.wait(5000)
```
tangkong, line 152: This is a bit of troublesome thread usage.  If python GC's this before the thread's run-loop is finished/broken, we'll segfault.  The wait time specified here is longer than we need, but the exact time we need depends on the number of pvs being polled and the specified loop period.  

I'm getting pulled in other directions so I might leave this as a followup

ZLLentz: I'm not sure I understand what's breaking here, but my small suggestion is to make adjustments in the `LivePVTableModel` class because accessing the thread from here seems like a breakdown of abstraction. The `CollectionBuilderPage` shouldn't need to manage the `LivePVTableModel`'s thread so closely.

superscore/widgets/page/collection_builder.py
```diff
@@ -0,0 +1,236 @@
...
+        # add model to two table views
+        logger.debug(f"Creating new model with {len(self.sub_pvs)} parameters "
+                     f"and {len(self.sub_colls)} collections")
+        self.pv_model = LivePVTableModel(entries=self.sub_pvs, client=self.client)
+        self.coll_model = NestableTableModel(entries=self.sub_colls)
+        self.sub_pv_table_view.setModel(self.pv_model)
+
+        # TODO: un-hard code this once there is a better way of managing columns
+        # Potentially dealing with columns that have moved
+        for i in [1, 4, 6]:
```
ZLLentz, line 163: I would feel less bad about magic column numbers if they were enums or named constants.

tangkong: Yeah, I do intend to make this less heinous.  I have some features I want to add to these tables and that will change how we interface with showing/hiding columns.

tangkong: For some reason I thought this was for the nestable table.  I'll use the enum members for now and refactor it later

ZLLentz:
I came back after lunch intending to continue my review but I don't see anything else major. I like the concept of that searchable/filterable combobox.

superscore/widgets/page/collection_builder.py
tangkong: I ended up addressing this, as I was going to work on the follow up later anyway

tangkong:
This should now close #82

superscore/widgets/views.py
```diff
@@ -593,19 +612,28 @@ def start_polling(self) -> None:
...
+    def stop_polling(self, wait_time: float = 0.0) -> None:
+        """
+        stop the polling thread, and mark it as stopped.
+        wait time in ms
+        """
+        if self._poll_thread is None or not self._poll_thread.isRunning():
             return
 
+        logger.debug(f"stopping and de-referencing thread @ {id(self._poll_thread)}")
+        # does not remove reference to avoid premature python garbage collection
```
ZLLentz, line 630: I don't really understand what the difference is between this and the previous but if the segfaults are gone great work

ZLLentz (approved):

superscore/widgets/views.py
tangkong: The diff here makes this comment look useless, but it's more of a reminder not to dereference `self._poll_thread`.  I went through a couple debugging iterations and notably atef does this

tangkong:
I identified a few followups, but for now I will merge this.  It's functional enough for us to move forward, I think.  And I'm going to want to use the table features in the next work

PR #89, opened 2024-09-27
BUG: close tabs in tab widget when main window is closed

### LINKED ISSUE #88: BUG: LivePVTableModel polling isn't stopped before deletion when main window is closed
## Current Behavior

Closing the main window while a tab with a `LivePVTableModel` is open causes this error:
```
QThread: Destroyed while thread is still running
Aborted (core dumped)
```
## Expected Behavior

Tabs with `LivePVTableModel` should stop their polling threads before closing to ensure clean shutdown.

1. `superscore ui`
2. Open collection builder page e.g. File -> New -> Collection
3. Close main window by clicking top-right `X` 

## Suggested Solution

I've tried implementing `closeEvent` in `window.py::Window` to close all tabs in tab widget, but this event doesn't seem to trigger.  `closeEvent` is also implemented in the collection builder page, but closing the main window doesn't seem to send close events to all widgets.

There are flags like `DeleteOnClose` and signals like `destroyed` that may be more reliable, but this reliability comes at bluntness, and closing seems like the preferred way to handle clean up.

shilorigins:
## Description

Fix bug where `LivePVTableModel` polling threads aren't cleaned up when the main window is closed.

Closes #88 
## Motivation and Context

See #88 
## How Has This Been Tested?

Interactively
## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/89-tab_widget_cleanup.rst
+++ b/docs/source/upcoming_release_notes/89-tab_widget_cleanup.rst
@@ -1,0 +1,22 @@
+89 tab widget cleanup
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- Main window closes tab widget tabs when closed, so they can do their clean up
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- shilorigins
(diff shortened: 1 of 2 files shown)


tangkong (approved):
While (har) I like the guarantee that a for loop will terminate, this should be sufficient to get all the tabs.  If the tab removal process ever somehow doesn't decrement the count, we'll only have ourselves to blame

ZLLentz (approved):

PR #93, opened 2024-10-15
ENH: add demo mode

shilorigins:
## Description

Add demo command.  This command creates a backend and launches an IOC using data pulled from fixtures before launching the UI, so that the app can be tested interactively with one pre-configured command.
## Motivation and Context

By using a dedicated config, starting from a consistent state, and handling IOC creation and clean-up, this command makes it simpler to interactively test the app.

Fixtures can be selected by adding the function name to the space-separated `fixtures` var under `[demo]` in `tests/demo.cfg`. Fixtures must be defined in `tests/conftest.py` (at least for now), and can return a single `Entry` or an iterable of `Entry`s.

To get this working, `Entry` and backend `get_entry` convert string uuids to proper UUID instances, because the mix of `str` and `UUID` types were causing duplicate entries to be considered distinct.  Additionally, new filestore db files are created as needed rather than requiring an initialization flag, and a bug where initialization assumes a valid `FilestoreBackend._root` has been resolved. 
## How Has This Been Tested?

Interactively.  No new tests have been added, but the existing test suite passes.

There! ^

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/93-add_demo_mode.rst
+++ b/docs/source/upcoming_release_notes/93-add_demo_mode.rst
@@ -1,0 +1,26 @@
+93 add demo mode
+#################
+
+API Breaks
+----------
+- `FilestoreBackend.__init__` does not accept `initialize` arg
+- `FilestoreBackend.store` does not accept any args
+
+Features
+--------
+- `demo` subcommand that launches the UI with a backend and IOC populated with PV
+  data from selectable fixtures
+- `ui.py::main` accepts an existing `Client` as an arg
+
+Bugfixes
+--------
+- `_root` was `None` during `FilestoreBackend` file initialization, causing `TypeError`
+- `IOCFactory.prepare_attrs` now strips attr names of all non-alphanumeric characters
+
+Maintenance
+-----------
+- Improve UUID / str interchangeability
+
+Contributors
+------------
+- shilorigins


tangkong (commented):
This looks great to me, just a couple little things, but I have no major qualms with this.

superscore/tests/demo.cfg
```diff
@@ -0,0 +1,10 @@
+[backend]
+type = filestore
+path = ./db/demo.json
```
tangkong, line 3: This will make a filestore backend and save it to the library folder.  We clean it up with each call, but if multiple people open up the demo they'll end up clobbering each other (I think)

Should this be the in-memory test-backend?

ZLLentz: Maybe we can use the user's home area instead so they can have some persistent demo data

tangkong: Building off of that, add a path argument to the arg parser so the user can place this demo wherever they want?

shilorigins: Multiple instances will clobber each other, true. I wanted it to be a filestore so the data persists after the run, but using the user's home would avoid clobbering. I figured a user would modify the config if they wanted to change the db file location.

tangkong: I was envisioning a central installation of `superscore`, where the user doesn't actually have access to the library files.  at least on the LCLS side, we have once central environment that people start apps from (and don't have write access to)

I suppose one could argue that if people are building their own config then we're past the point of demo-ing already

shilorigins: I was only thinking of us using the mode for testing and demos, not actual users.  I can make it more robust if we think users might find it useful as well.

tangkong: I think this will be too useful not to advertise to users who want to play around with it.  😆

superscore/bin/demo.py
```diff
@@ -0,0 +1,48 @@
...
+    except FileNotFoundError:
+        pass
+    parser = configparser.ConfigParser()
+    parser.read(DEMO_CONFIG)
+    filled = []  # IOCFactory needs the Entries with data
+    for fixture_name in parser.get("demo", "fixtures").split():
+        fixture = getattr(superscore.tests.conftest, fixture_name)
+        data = fixture()
+        # fixtures can return single Entries or iterables of Entries
+        entries = (data,) if isinstance(data, Entry) else data
```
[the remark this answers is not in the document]
shilorigins: I added an explicit error message if the return type isn't right.

I looked at supporting backends and clients, but since `pytest` doesn't want us to call fixtures directly, it broke the fixture dependencies. Getting them to work would take some reorganization and we'd likely lose a lot of convenience, but I can spend more time on it in the future if we would find that feature useful.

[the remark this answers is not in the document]
tangkong: If it's more than a quick fix we can think about it later.  Fixtures really shouldn't be called outside of pytest, there's a lot of dark magic involved in setup and teardown that makes them special.  But I suppose until we get burnt by making fixtures rely on normal functions, we can run with this usage mode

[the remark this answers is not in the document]
shilorigins: I think it's okay as long as the "fixture" doesn't have any args, like for functions that just instantiate a bunch of `Entry`s.  Outside of that there be dragons

superscore/tests/ioc/ioc_factory.py
```diff
@@ -66,6 +66,6 @@ def prepare_attrs(entries: Iterable[Entry]) -> Mapping[str, pvproperty]:
         for entry in pvs:
             value = entry.data if isinstance(entry, (Setpoint, Readback)) else None
             pv = pvproperty(name=entry.pv_name, doc=entry.description, value=value, dtype=dbr.DBR_STRING if isinstance(entry.data, str) else None)
-            attr = "".join([c.lower() for c in entry.pv_name if c != ':'])
+            attr = "".join([c.lower() for c in entry.pv_name if c not in [':', '.']])
```
tangkong, line 69: (Something I should have spotted earlier)

I think technically there are no restrictions on PV names, only our loose conventions.  That means that to be complete we should really be stripping any character that isn't valid in python variables.  

I think the easiest check is to require all characters be alphanumeric + "_".  Since the user never accesses these attributes, I think it's worth being more strict in exchange for avoiding weird edge cases.  

But also this is basically a nitpick.

tangkong (dismissed):
I'm on board, excited to use this in testing! 🎉

tangkong:
Just a little thing I noticed, I have no real substantive objections to this 👍

superscore/tests/ioc/ioc_factory.py
shilorigins: Good point, might as well get it while we're here.

shilorigins:
Just cleaned up the commits and finished adding the last suggestions, thanks for the reveiws!

tangkong (approved):
Looks good to me 👍

PR #94, opened 2024-10-15
ENH: add snapshot view / compare / restore page (only view functional)

shilorigins:
## Description

Add restore page template and logic
Add ability to open restore page from root tree `QTreeView`
Increase default size of the main window
Fix bug where closing a `LivePVTableModel` sometimes causes an error trying to access deleted data

## Motivation and Context

This page can currently show all leaf `Entry`s in a snapshot and can compare stored values to live values.  The same page is intended to be able to compare values from two snapshots and to restore saved values, but both of those capabilities are postponed for later PRs.

The AD control system has a convention where middle clicking on a UI element shows the underlying PV and copies it to the clipboard, so I added middle-click-to-copy as a partial implementation and convenience feature.  We can fill out the PV-display-and-copy later if desired. 

~Double-clicking an entry on the tree view opens the entry page and no longer expands the tree item because I found that use case intuitive, and items can still be expanded by clicking on the items' arrows.  Happy to alter this functionality if doing so would be more usable.~
Since the `Snapshot` entry page is now associated with the tree view's double-click, I added an action to open the restore page to the tree view's context menu.

I increased the size of the main window to better show the data-dense tables without having to manually expand the window every time.  I didn't optimize the new size, but it works well enough on my dev monitor and laptop.

## How Has This Been Tested?

Interactively.
## Screenshots:
Page when a snapshot is opened:
Screenshot 2024-10-23 at 15 56 24

Page when `Compare to Live` button is clicked:
Screenshot 2024-10-23 at 15 56 36

## Pre-merge checklist

### DIFF
--- a/superscore/bin/ui.py
+++ b/superscore/bin/ui.py
@@ -10,6 +10,9 @@ from qtpy.QtWidgets import QApplication
 from superscore.client import Client
 from superscore.widgets.window import Window
 
+DEFAULT_WIDTH = 1400
+DEFAULT_HEIGHT = 800
+
 
 def build_arg_parser(argparser=None):
     if argparser is None:
@@ -22,5 +25,11 @@ def main(*args, client: Optional[Client] = None, **kwargs):
     app = QApplication(sys.argv)
     main_window = Window(client=client)
 
+    primary_screen = app.screens()[0]
+    center = primary_screen.geometry().center()
+    # move window rather creating a QRect because we want to include the frame geometry
+    main_window.setGeometry(0, 0, DEFAULT_WIDTH, DEFAULT_HEIGHT)
+    delta = main_window.geometry().center()
+    main_window.move(center - delta)
     main_window.show()
     app.exec()
--- a/superscore/tests/conftest.py
+++ b/superscore/tests/conftest.py
@@ -12,7 +12,7 @@ from superscore.client import Client
 from superscore.control_layers._base_shim import _BaseShim
 from superscore.control_layers.core import ControlLayer
 from superscore.model import (Collection, Parameter, Readback, Root, Setpoint,
-                              Snapshot)
+                              Severity, Snapshot, Status)
 from superscore.tests.ioc import IOCFactory
 
 
@@ -311,6 +311,8 @@ def linac_data():
         pv_name=lasr_gunb_pv1.pv_name,
         description=lasr_gunb_pv1.description,
         data="Off",
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     now = lasr_gunb_value1.creation_time
@@ -321,6 +323,8 @@ def linac_data():
         description=lasr_gunb_pv2.description,
         creation_time=now,
         data=5,
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     lasr_gunb_snapshot = Snapshot(
@@ -340,6 +344,8 @@ def linac_data():
         description=mgnt_gunb_pv.description,
         creation_time=now,
         data=True,
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     mgnt_gunb_snapshot = Snapshot(
@@ -358,6 +364,8 @@ def linac_data():
         description=vac_gunb_pv1.description,
         creation_time=now,
         data="Ion Pump",
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     vac_gunb_value2 = Setpoint(
@@ -366,6 +374,8 @@ def linac_data():
         description=vac_gunb_pv2.description,
         creation_time=now,
         data=False,
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     vac_gunb_snapshot = Snapshot(
@@ -397,6 +407,8 @@ def linac_data():
         description=vac_l0b_pv.description,
         creation_time=now,
         data=-10,
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     vac_l0b_snapshot = Snapshot(
@@ -425,6 +437,8 @@ def linac_data():
         description=vac_bsy_pv.description,
         creation_time=now,
         data="",
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     vac_bsy_snapshot = Snapshot(
@@ -465,6 +479,8 @@ def linac_data():
         description=vac_li10_pv.description,
         creation_time=now,
         data=.25,
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     vac_li10_snapshot = Snapshot(
@@ -493,6 +509,8 @@ def linac_data():
         description=lasr_in10_pv.description,
         creation_time=now,
         data=645.26,
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     lasr_in10_snapshot = Snapshot(
@@ -532,6 +550,8 @@ def linac_data():
         description=lasr_in20_pv.description,
         creation_time=now,
         data=0,
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     lasr_in20_snapshot = Snapshot(
@@ -560,6 +580,8 @@ def linac_data():
         description=vac_li21_pv.description,
         creation_time=now,
         data=0.0,
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
     )
 
     vac_li21_setpoint = Setpoint(
@@ -568,6 +590,8 @@ def linac_data():
         description=vac_li21_pv.description,
         creation_time=now,
         data=5.0,
+        status=Status.NO_ALARM,
+        severity=Severity.NO_ALARM,
         readback=vac_li21_readback,
     )
 
--- a/superscore/tests/test_page.py
+++ b/superscore/tests/test_page.py
@@ -14,6 +14,7 @@ from superscore.widgets.page.collection_builder import CollectionBuilderPage
 from superscore.widgets.page.entry import (BaseParameterPage, CollectionPage,
                                            ParameterPage, ReadbackPage,
                                            SetpointPage, SnapshotPage)
+from superscore.widgets.page.restore import RestorePage
 from superscore.widgets.page.search import SearchPage
 
 
@@ -81,6 +82,14 @@ def collection_builder_page(qtbot: QtBot, sample_client: Client):
     qtbot.waitUntil(lambda: page.sub_pv_table_view._model._poll_thread.isFinished())
 
 
+@pytest.fixture
+def restore_page(qtbot: QtBot, sample_client: Client, simple_snapshot: Snapshot):
+    page = RestorePage(data=simple_snapshot, client=sample_client)
+    qtbot.addWidget(page)
+    yield page
+    page.close()
+
+
 @pytest.mark.parametrize(
     'page',
     [
@@ -91,6 +100,7 @@ def collection_builder_page(qtbot: QtBot, sample_client: Client):
         "snapshot_page",
         "search_page",
         "collection_builder_page",
+        "restore_page",
     ]
 )
 def test_page_smoke(page: str, request: pytest.FixtureRequest):
@@ -218,3 +228,16 @@ def test_stored_widget_swap(
         qtbot.waitUntil(
             lambda: isinstance(page.value_stored_widget, expected_widget),
         )
+
+
+def test_restore_page_toggle_live(restore_page):
+    tableView = restore_page.tableView
+    live_columns = tableView.live_headers
+    toggle_live_button = restore_page.compareLiveButton
+
+    toggle_live_button.click()
+    assert tableView._model._poll_thread.running
+    assert all((not tableView.isColumnHidden(column) for column in live_columns))
+
+    toggle_live_button.click()
+    assert all((tableView.isColumnHidden(column) for column in live_columns))
--- a/superscore/tests/test_page.py
+++ b/superscore/tests/test_page.py
@@ -230,14 +230,14 @@ def test_stored_widget_swap(
         )
 
 
-def test_restore_page_toggle_live(restore_page):
+def test_restore_page_toggle_live(qtbot: QtBot, restore_page):
     tableView = restore_page.tableView
     live_columns = tableView.live_headers
     toggle_live_button = restore_page.compareLiveButton
 
     toggle_live_button.click()
-    assert tableView._model._poll_thread.running
-    assert all((not tableView.isColumnHidden(column) for column in live_columns))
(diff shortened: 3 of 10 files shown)


tangkong:
I grabbed the changes and they look pretty good.  No major bugs, though there are a limited number of interaction points thus far.  From interactive testing:
* the open button is not wired, since there was no open_page_slot provided.  If the intent is for this to not be supported, we should also hide the open column

I'll go through the code in a bit more detail now

shilorigins:
I was just adding some (any) tests before requesting review. Is there a setpoint / readback page to connect to `open_page_slot` yet?

superscore/widgets/page/restore.py
```diff
@@ -0,0 +1,84 @@
...
+from superscore.client import Client
+from superscore.model import Snapshot
+from superscore.widgets.core import Display
+from superscore.widgets.views import LivePVHeader, LivePVTableView
+
+logger = logging.getLogger(__name__)
+
+
+class Restore(Display, QtWidgets.QWidget):
+    """
```
tangkong, line 18: Are we missing a docstring? here?

superscore/widgets/page/restore.py
```diff
@@ -0,0 +1,84 @@
...
+
+from superscore.client import Client
+from superscore.model import Snapshot
+from superscore.widgets.core import Display
+from superscore.widgets.views import LivePVHeader, LivePVTableView
+
+logger = logging.getLogger(__name__)
+
+
+class Restore(Display, QtWidgets.QWidget):
```
tangkong, line 17: Nitpick: I like calling these `Pages` (aka `RestorePage`), just to make it abundandly clear that this isn't itself the restore action or procedure.

superscore/widgets/page/restore.py
```diff
@@ -0,0 +1,84 @@
...
+
+    headerFrame: QtWidgets.QFrame
+    primarySnapshotLabel: QtWidgets.QLabel
+    primarySnapshotTitle: QtWidgets.QLabel
+    secondarySnapshotLabel: QtWidgets.QLabel
+    secondarySnapshotTitle: QtWidgets.QLabel
+    compareLiveButton: QtWidgets.QPushButton
+    compareSnapshotButton: QtWidgets.QPushButton
+
+    tableView: LivePVTableView
```
tangkong, line 31:
```suggestion
    tableView: SnapshotTableView
```

superscore/widgets/page/restore.py
```diff
@@ -0,0 +1,84 @@
...
+        self.tableView.close()
+        super().closeEvent(a0)
+
+
+class SnapshotTableView(LivePVTableView):
+    """"""
+    live_headers = {LivePVHeader.LIVE_VALUE, LivePVHeader.LIVE_STATUS, LivePVHeader.LIVE_SEVERITY}
+
+    def gather_sub_entries(self) -> None:
+        self.sub_entries = self.client._gather_leaves(self.data)
```
tangkong, line 79: I did sort of expect us to be doing the gathering outside the LivePVTableView, then passing the entries to set_data, but this works too I suppose.

superscore/widgets/page/restore.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def closeEvent(self, a0: QCloseEvent) -> None:
```
tangkong: I'm realizing this should probably be in `LivePVTableView`, rather than in the pages that contain it (my Collection builder page, for example) or in subclasses.

superscore/widgets/views.py
```diff
@@ -1154,6 +1153,19 @@ def set_editable(self, column: int, is_editable: bool) -> None:
             return
         self._model.set_editable(column, is_editable)
 
+    def mousePressEvent(self, event: QtGui.QMouseEvent) -> None:
+        if event.button() == QtCore.Qt.MiddleButton:
```
tangkong, line 1157: If I had mouse on hand I'd test this 😆

superscore/widgets/window.py
```diff
@@ -42,6 +47,8 @@ def __init__(self, *args, client: Optional[Client] = None, **kwargs):
 
         self._partial_slots = []
 
+        self.resize(DEFAULT_WIDTH, DEFAULT_HEIGHT)
+        self.move(self.screen().geometry().center() - self.frameGeometry().center())
```
tangkong, line 51: For me this centers the window in-between my two displays.  We may need to pick a display before centering

superscore/widgets/window.py
```diff
@@ -107,8 +116,11 @@ def open_page(self, entry: Entry) -> None:
         try:
             page = PAGE_MAP[type(entry)]
         except KeyError:
-            logger.debug(f'No page widget for {type(entry)}, cannot open in tab')
-            return
+            if isinstance(entry, Snapshot):
+                page = partial(Restore, client=self.client)
```
tangkong, line 120: I'll move this to PAGE_MAP in a followup PR (or when I rebase my entry pages).  I think we should be passing the client through to all the pages anyway, and I've changed this in one of my branches already

tangkong:
> I was just adding some (any) tests before requesting review. Is there a setpoint / readback page to connect to `open_page_slot` yet?

whoops sorry I jumped the gun a bit.  I have pages in #95, and have been adding tests there as well (though I haven't managed to finish and push them yet 💀 )

superscore/widgets/page/restore.py
shilorigins: Good thing I moved it just before you posted your review haha

superscore/widgets/window.py
shilorigins: Hmm interesting.  I'm using NoMachine, so it centers on one display despite my having two monitors.  I'll find a way to reproduce this on my setup.

superscore/widgets/page/restore.py
shilorigins: `BaseDataTableView.set_data` calls `maybe_setup_model`, which calls `gather_sub_entries`.  Since I want to flatten the snapshot into all of its leaves regardless of depth (at least for now), overwriting `gather_sub_entries` seemed the most natural way.

superscore/widgets/page/restore.py
shilorigins: Indeed, I've added a short one.

superscore/widgets/page/restore.py
shilorigins: Nice catch

superscore/widgets/page/restore.py
tangkong: yep, many paths through the forest.  I have no problem with this, aside from a very personal desire to avoid subclassing when possible.  It is quite a clean approach 👍

superscore/widgets/window.py
tangkong: I have an extended display, and when I start `superscore demo` it opens the window at the boundary between my two displays

Screenshot 2024-10-25 at 11 48 48 AM (2)

image

tangkong: With extended display and a single display, this is always the result of the screen calculation:
```python
In [1]: page.screen().geometry().center() - page.frameGeometry().center()
Out[1]: PyQt5.QtCore.QPoint(0, -11)
```

tangkong: With extended display:
```python
In [3]: page.screen().geometry()
Out[3]: PyQt5.QtCore.QRect(0, 0, 3648, 1080)
```

with single display:
```python
In [2]: page.screen().geometry()
Out[2]: PyQt5.QtCore.QRect(0, 0, 1728, 1080)
```

So the `page.screen()` sees my two displays as one huge one, then correctly places the window in the center of them

shilorigins: I'm also using a main and extended display 🤔

tangkong: I suppose that if it's an "extended display" setup, there's no way for qt to really know there are two monitors involved.  If there were multiple monitors things like `QDesktopWidget().screenCount()` could help us out, but in my case that will still return 1

On the photon side our hutch control rooms sometimes involve 8+ monitors, I'd imagine it's a similar story in ACR?

Is it worth trying to make assumptions about the monitor setup across these different use cases? (local vs prod vs ctrl room).  Should we maybe just leave it in the default location? (top left?) 

I'm not sure I have a strong opinion yet, but I do think extended displays are not an unusual use case

shilorigins: Maybe we leave it in the default spot (upper left) for now, since it seems like it'll take some effort to sort out.

What's a little strange is that the window opened in the middle of the screen before I resized it, so whatever was putting it there was sorting out the monitor issue.

shilorigins: I pushed a commit where I select the primary screen for centering.  Let me know if you still have an issue, and I'll just reset to the default logic for now.

I experimented in my env a bit, and can confirm that `app.screens()` reports both of my two monitors as separate and their correct sizes.  We should one day investigate why our envs are showing as different.

tangkong: this works slight better for me, though `app.screens()` still reports 1 for my setup.   

After thinking/investigating this a little more, I actually think this is due to x-forwarding.  If I'm fully local, `app.screens()` reports 2, but if I'm logged into our controls servers and forwarding screens qt sees only one.

This is probably fine to leave as you've implemented here.  I don't think this is a huge issue either way.  For the record we tend to do screen movement/rearrangement outside of the application, choosing to have it in the scripts/entrypoints that call the applications.  But also it's fine for superscore to do a best effort first pass at this

shilorigins:
Since a branch merged into master replaced the default `Snapshot` page, I added a context menu action to open the restore page. We can link the pages better in the future, but I think we should discuss the intended page flow first.

superscore/bin/ui.py
```diff
@@ -10,6 +10,9 @@
 from superscore.client import Client
 from superscore.widgets.window import Window
 
+DEFAULT_WIDTH = 1400
+DEFAULT_HEIGHT = 800
```
tangkong, line 14: nitpick: I'm always wary of setting raw defaults for things like sizes, since peoples' display setups vary so much.  There's probably a way to set this relative to a user's total screen real estate 🤔 .  

I also won't argue that superscore needs more space than qt was originally providing it.  So I don't object strongly to this

superscore/tests/conftest.py
```diff
@@ -311,6 +311,8 @@ def linac_data():
         pv_name=lasr_gunb_pv1.pv_name,
         description=lasr_gunb_pv1.description,
         data="Off",
+        status=Status.NO_ALARM,
```
tangkong, line 314: A shame you had to specify all these, but I do think our defaults still make sense.

superscore/tests/test_page.py
```diff
@@ -218,3 +228,16 @@ def simple_coll_return_vals(pv_name: str):
...
+
+
+def test_restore_page_toggle_live(restore_page):
+    tableView = restore_page.tableView
+    live_columns = tableView.live_headers
+    toggle_live_button = restore_page.compareLiveButton
+
+    toggle_live_button.click()
+    assert tableView._model._poll_thread.running
+    assert all((not tableView.isColumnHidden(column) for column in live_columns))
```
tangkong, line 240: I would wrap this assert in a `qtbot.waitUntil` or something similar.  Whenever testing something involving a callback / signal, there's always a chance we hit the asserts before the event loop has processed things.  This also guards against us adding more callbacks (with potentially blocking code), without becoming too much more brittle.

It probably never errors for you locally, but it's only a matter of time until some shenanigans happen in CI

superscore/widgets/page/restore.py
```diff
@@ -0,0 +1,121 @@
...
+        for live_header in self.live_headers:
+            self.setColumnHidden(live_header, not self._is_live)
+        if self._is_live:
+            self.turnOnLive.emit()
+        else:
+            self.turnOffLive.emit()
+
+    @QtCore.Slot()
+    def toggle_live(self):
+        self.set_live(not self._is_live)
```
tangkong, line 54: I think pulling this up into LivePVTableView would be really nice, but is definitely for a different PR

superscore/widgets/page/restore.py
```diff
@@ -0,0 +1,121 @@
...
+    """"""
+    def data(self, index: QtCore.QModelIndex, role: int):
+        if role == QtCore.Qt.TextAlignmentRole:
+            return QtCore.Qt.AlignCenter
+        else:
+            return super().data(index, role)
+
+
+class SnapshotTableView(LivePVTableView):
+    """"""
```
tangkong, line 28: Did some docstrings get lost here?  (And for `LiveSnapshotProxyModel`?)

superscore/widgets/page/restore.py
```diff
@@ -0,0 +1,121 @@
...
+from superscore.widgets.views import (LivePVHeader, LivePVTableModel,
+                                      LivePVTableView)
+
+logger = logging.getLogger(__name__)
+
+
+class LiveSnapshotProxyModel(LivePVTableModel):
+    """"""
+    def data(self, index: QtCore.QModelIndex, role: int):
+        if role == QtCore.Qt.TextAlignmentRole:
```
tangkong, line 21: We discussed this offline, but handling text alignment is also something possibly worth pulling up into `LivePVTableModel` in a future effort

tangkong (commented):
I think this looks pretty great code-wise, you've come up to speed on qt pretty quickly I think!

I'm holding off on approving until I can take it for a spin.  I'm hoping to get to that tomorrow

superscore/tests/conftest.py
shilorigins: I agree.  I couldn't come up with a nice way to generate test snapshots without taking too much time thinking about it, but we might want to revisit if we need to generate more.  The best idea I had was mocking `uuid4()` and `ControlLayer.get` with consistent, pre-selected data and then using `Client.snap`.

superscore/bin/ui.py
shilorigins: I agree, but I don't know how to manage screen details without researching it a bit.  Like how your setup reports one combined screen while mine reports two screens.  I suppose we could do percent of primary screen with a maximum absolute size?

superscore/widgets/page/restore.py
shilorigins: Indeed, I lost them in the other PR 😅

superscore/tests/test_page.py
shilorigins: Thanks for pointing this out!

superscore/bin/ui.py
tangkong: Yea, there's a few ideas but we can address these in more detail later

superscore/tests/conftest.py
tangkong: Related but not directly applicable: I think if we can set up some way of selecting a client/backend-data combination using our fixtures, that would be ideal.  We already have a variety of datasets, but our `sample_client` only used the filestore json.  

You could also take a snapshot of your controls system and save it in a filestore-backend, then we could add that as a backend to test with?

tangkong (approved):
LGTM.  I took this for a quick spin and things worked as advertised.  The only additional thing I noticed was the context menu option that opened the restore page was named "inspect", but we need to clear up the page flow anyway

Issue #68, opened 2024-07-23
GUI: Build Value (Readback / Setpoint) page widget


tangkong:

tangkong:
closed by #95

Issue #67, opened 2024-07-23
GUI: Build Snapshot page view widget


tangkong:
This view should allow users to inspect and explore the components of a Snapshot, including metadata, child Entries, and their values. May be the foundation for a view comparing two snapshots (#48)

tangkong:
closed by #95

Issue #66, opened 2024-07-23
GUI: Build Parameter page view widget


tangkong:

tangkong:
closed by #95

Issue #65, opened 2024-07-23
GUI: Build Collection page view widget


tangkong:
This view should allow users to inspect and explore the components of a Collection, including metadata and child Entries.

tangkong:
Closed by #95

PR #95, opened 2024-10-22
GUI: Adds Entry pages (BaseParameterPage, NestablePage)

tangkong:
## Description
Adds Pages for viewing and editing details for the other entry types.

- Adds `BaseParameterPage` which dynamically shows and hides widgets based on which fields are present (Single PV Entries)
- adds BusyCursorThread, for simple blocking work
- slightly modify the signature of open_page_slot, but make a type alias for it so we get it right in the future
- add save-button logic (save to client, disable/hide if not editable)
- Add editable flag/toggle
- complete nestable pages (Collection view, Snapshot view)
- more tests

This is currently based on top of #96 , once that's in I'll rebase one last time

## Motivation and Context
#65 
#66 
#67 
#68 

## How Has This Been Tested?
I'll add tests I swear

This PR

A readback (has tolerances, no rbv)
image

A setpoint (no tolerances, has rbv)
image

Widgets update based on pv type
image

## Pre-merge checklist


superscore/widgets/__init__.py
```diff
@@ -1,10 +1,12 @@
 def _get_icon_map():
     # do not pollute namespace
-    from superscore.model import Collection, Readback, Setpoint, Snapshot
+    from superscore.model import (Collection, Parameter, Readback, Setpoint,
+                                  Snapshot)
 
     # a list of qtawesome icon names
     icon_map = {
         Collection: 'mdi.file-document-multiple',
+        Parameter: 'msc.symbol-field',
```
ZLLentz, line 9: Did you try a few symbol variants? This one is somewhat unintuitive to me. Did something like `mdi.file-document` not work for some design reason?

superscore/widgets/page/entry.py
```diff
@@ -1,36 +1,382 @@
...
+            self.bridge.abs_tolerance
+        except AttributeError:
+            self.tol_widget.hide()
+        else:
+            self.abs_tol_spinbox.setValue(self.data.abs_tolerance or 0.0)
+            self.rel_tol_spinbox.setValue(self.data.rel_tolerance or 0.0)
+            self.update_tol_calc()
+
+            self.abs_tol_spinbox.valueChanged.connect(self.update_abs_tol)
+            self.rel_tol_spinbox.valueChanged.connect(self.update_rel_tol)
```
ZLLentz, line 224: I'm curious about the code structure here.
It seems like this section checks if the data bridge has a particular field available, and sets up the widget appropriately, and everything paths through this large code block.

Is this easier to maintain than the alternative, which would be implementing specific behavior in each subclass based on which fields we know are or are not present?

ZLLentz (dismissed):
I asked some questions but things look reasonable
I wasn't able to fully test myself because I couldn't get an editable entry page to appear from superscore's ui

superscore/widgets/__init__.py
tangkong: I admittedly didn't think super hard about this, my mind was thinking of Parameters as a "building block" of sorts.  But a single `mdi.file-document` combining into a Collection (file-document-multiple) makes a lot of sense, I was blind to the obvious thing

superscore/widgets/page/entry.py
tangkong: At the point where I'd end up hiding and/or setting up the same widgets in multiple classes, I just thought it might make sense to do it all in one class.  For example why set up the tolerance fields in both `Parameter` and `Readback`, or why hide the timeout twice in both `Setpoint` and `Parameter`?  I'd end up just copying the same code to another subclass.  

This next statement is more of a stretch, but it does also let us change the schema and not worry about having to change the widgets.  (I do also think we're unlikely to do this, in all fairness)

Admittedly inspecting the data bridge is a step of indirection, but we do expect it to mirror the contained dataclass.

superscore/widgets/__init__.py
tangkong: Seeking other opinions here: is this too similar? 
image

ZLLentz: It's certainly a bit closer than I'd like but I'm not sure
There's also an outline variant we could try

tangkong:
A quick icon change, then I think we can get this in

ZLLentz (approved):

PR #96, opened 2024-10-22
FIX: actually fill child entries in shared table views

tangkong:
## Description
Actually fill child entries if they are uuids

## Motivation and Context
Originally observed as a bug where entries would not be loaded into tables when you'd expect them to.  Also modifies the parent dataclass, replacing UUIDs with an Entry.  This does not recurse.

## How Has This Been Tested?
Interactively, and some tests were added

This PR.

## Pre-merge checklist

### DIFF
--- a/superscore/tests/test_views.py
+++ b/superscore/tests/test_views.py
@@ -1,17 +1,20 @@
 import copy
 from typing import Any
 from unittest.mock import MagicMock
+from uuid import UUID
 
 import apischema
 import pytest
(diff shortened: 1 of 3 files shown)


superscore/tests/test_views.py
ZLLentz, line 1: It's worth adding that you wrote unit tests to the PR description's testing section

ZLLentz (approved):

PR #97, opened 2024-11-06
ENH: add restore functionality to snapshot restore page

shilorigins:
## Description

* Add "restore" button to restore page
* Button opens dialog with all PVs and values in a table, plus accept button
  * PVs can be removed
  * Accept button writes all values in the table to the EPICS system
  * Accept button also turns on live data tracking, as a way of tracking restore progress
## Motivation and Context

Restoring snapshots and other sets of values are a core feature of the application.

A dialog represents an action being configured and then activated, so I've used it to coordinate the restore action for now.  Alternatives include filtering the `RestorePage` main PV table, but this could muddy the app's state in the mind of the user unless it's done well.

PVs can be removed from the restore dialog, but currently not added in.  This covers cases where a whole or partial `Snapshot` needs to be restored, but not multiple `Snapshot`s or cases where the user accidentally removes a PV.  This is an acceptable first pass while the dialog / restore page continue to be developed.

An alternative is to replace the "Remove" push button with a radio button that marks the PV as active or inactive: inactive PVs would stay in the dialog table but only active PVs would be included in the restore.  This is an easy way to make PVs re-includable, but may be visually busy for large snapshots.  This method also lends itself to being used in the main table, obviating the need for a dialog at all.
## How Has This Been Tested?

Interactively:
1. run `superscore demo`
2. right click "Accelerator Directorate" snapshot in the tree view and click "inspect values"
3. click "Turn on Live Data"
4. click "Restore" button
5. click "Remove" button to un-stage PVs if desired
6. click dialog's "Restore" button
7. observe values changing in the `RestorePage` PV table

A second "linac" snapshot with different values has been added for comparing to the original snapshot.

I also added two test cases in `test_page.py`.
## Screenshots (if appropriate):
Restore page before restore:
Screenshot 2024-11-07 at 09 39 52

Restore dialog:
Screenshot 2024-11-08 at 11 23 05

Restore page after restore:
Screenshot 2024-11-07 at 09 41 39

## Pre-merge checklist

### DIFF
--- a/superscore/tests/ioc/ioc_factory.py
+++ b/superscore/tests/ioc/ioc_factory.py
@@ -1,11 +1,12 @@
 from multiprocessing import Process
-from typing import Iterable, Mapping, Union
+from typing import Iterable, Mapping
 
 from caproto.server import PVGroup, pvproperty
 from caproto.server import run as run_ioc
 from epicscorelibs.ca import dbr
 
-from superscore.model import Entry, Nestable, Parameter, Readback, Setpoint
+from superscore.client import Client
+from superscore.model import Entry, Readback, Setpoint
 
 
 class TempIOC(PVGroup):
@@ -32,36 +33,26 @@ class IOCFactory:
     Generates TempIOC subclasses bound to a set of PVs.
     """
     @staticmethod
-    def from_entries(entries: Iterable[Entry], **ioc_options) -> PVGroup:
+    def from_entries(entries: Iterable[Entry], client: Client, **ioc_options) -> PVGroup:
         """
         Defines and instantiates a TempIOC subclass containing all PVs reachable
         from entries.
         """
-        attrs = IOCFactory.prepare_attrs(entries)
+        attrs = IOCFactory.prepare_attrs(entries, client)
         IOC = type("IOC", (TempIOC,), attrs)
         return IOC
 
     @staticmethod
-    def collect_pvs(entries: Iterable[Entry]) -> Iterable[Union[Parameter, Setpoint, Readback]]:
-        """Returns a collection of all PVs reachable from entries"""
-        pvs = []
-        q = entries.copy()
-        while len(q) > 0:
-            entry = q.pop()
-            if isinstance(entry, Nestable):
-                q.extend(entry.children)
-            else:
-                pvs.append(entry)
-        return pvs
-
-    @staticmethod
-    def prepare_attrs(entries: Iterable[Entry]) -> Mapping[str, pvproperty]:
+    def prepare_attrs(entries: Iterable[Entry], client: Client) -> Mapping[str, pvproperty]:
         """
         Turns a collecton of PVs into a Mapping from attribute names to
         caproto.pvproperties. The mapping is suitable for passing into a type()
         call as the dict arg.
         """
-        pvs = IOCFactory.collect_pvs(entries)
+        pvs = []
+        for entry in entries:
+            leaves = client._gather_leaves(entry)
+            pvs.extend(leaves)
         attrs = {}
         for entry in pvs:
             value = entry.data if isinstance(entry, (Setpoint, Readback)) else None
--- a/superscore/tests/test_page.py
+++ b/superscore/tests/test_page.py
@@ -256,8 +256,14 @@ def test_restore_dialog_restore(
 
 def test_restore_dialog_remove_pv(mock_client: Client, simple_snapshot: Snapshot):
     dialog = RestoreDialog(mock_client, simple_snapshot)
-    assert dialog.tableWidget.rowCount() == len(simple_snapshot.children)
-    remove_button_column = 2
-    dialog.tableWidget.setCurrentCell(1, remove_button_column)
+    tableWidget = dialog.tableWidget
+    assert tableWidget.rowCount() == len(simple_snapshot.children)
+
+    PV_COLUMN = 0
+    REMOVE_BUTTON_COLUMN = 2
+    item_to_remove = tableWidget.item(1, PV_COLUMN)
+    tableWidget.setCurrentCell(1, REMOVE_BUTTON_COLUMN)
     dialog.delete_row()
-    assert dialog.tableWidget.rowCount() == len(simple_snapshot.children) - 1
+    assert tableWidget.rowCount() == len(simple_snapshot.children) - 1
+    items_left = [tableWidget.item(row, PV_COLUMN) for row in range(tableWidget.rowCount())]
+    assert item_to_remove not in items_left
--- a/superscore/tests/test_page.py
+++ b/superscore/tests/test_page.py
@@ -1,6 +1,6 @@
 """Largely smoke tests for various pages"""
 
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import pytest
 from pytestqt.qtbot import QtBot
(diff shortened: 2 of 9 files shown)


superscore/tests/ioc/ioc_factory.py
```diff
@@ -32,36 +33,26 @@ class IOCFactory:
...
-            entry = q.pop()
-            if isinstance(entry, Nestable):
-                q.extend(entry.children)
-            else:
-                pvs.append(entry)
-        return pvs
-
-    @staticmethod
-    def prepare_attrs(entries: Iterable[Entry]) -> Mapping[str, pvproperty]:
+    def prepare_attrs(entries: Iterable[Entry], client: Client) -> Mapping[str, pvproperty]:
```
tangkong, line 46: I'm not sure I'm on board with requiring the `IocFactory` to also be passed a client, when all it's doing is gathering leaf entries.  That means if we want to use this elsewhere, we need to construct a valid `Client` with a backend that holds the entries.

I know it's a bit repetitious, but here I think it could be justified

superscore/tests/test_page.py
```diff
@@ -241,3 +241,23 @@ def test_restore_page_toggle_live(qtbot: QtBot, restore_page):
...
+    assert put_mock.call_args.args == mock_client._gather_data(simple_snapshot)
+
+
+def test_restore_dialog_remove_pv(mock_client: Client, simple_snapshot: Snapshot):
+    dialog = RestoreDialog(mock_client, simple_snapshot)
+    assert dialog.tableWidget.rowCount() == len(simple_snapshot.children)
+    remove_button_column = 2
+    dialog.tableWidget.setCurrentCell(1, remove_button_column)
+    dialog.delete_row()
+    assert dialog.tableWidget.rowCount() == len(simple_snapshot.children) - 1
```
tangkong, line 263: I think it's also worth checking to make sure the row you think you deleted actually got deleted.  (as opposed to the last row being removed every time, or something equally silly)

superscore/tests/test_page.py
```diff
@@ -241,3 +241,23 @@ def test_restore_page_toggle_live(qtbot: QtBot, restore_page):
...
+
+@patch('superscore.control_layers.core.ControlLayer.put')
+def test_restore_dialog_restore(
+    put_mock,
+    mock_client: Client,
+    simple_snapshot: Snapshot,
+):
+    dialog = RestoreDialog(mock_client, simple_snapshot)
+    dialog.restore()
+    assert put_mock.call_args.args == mock_client._gather_data(simple_snapshot)
```
tangkong, line 254: I dig how simple this assertion turned out

superscore/widgets/page/restore.py
```diff
@@ -54,6 +54,52 @@ def toggle_live(self):
...
+            remove_item.clicked.connect(self.delete_row)
+            self.tableWidget.setCellWidget(row, 2, remove_item)
+
+        self.restoreButton.clicked.connect(self.restore)
+        self.cancelButton.clicked.connect(self.deleteLater)
+
+    def restore(self):
+        ephemeral_snapshot = Snapshot(children=self.entries)
+        self.client.apply(ephemeral_snapshot)
+        self.deleteLater()
```
tangkong, line 95: Did you find you needed to make this call?  Normally I just call `.close()` and let the built-in garbage collection do the work of scheduling deletion

superscore/widgets/page/restore.py
```diff
@@ -110,12 +157,19 @@ def __init__(
...
+        self.restoreButton.clicked.connect(self.launch_dialog)
+
     def set_live(self, is_live: bool):
         self.secondarySnapshotLabel.setVisible(is_live)
         self.secondarySnapshotTitle.setText("Live Data")
         self.secondarySnapshotTitle.setVisible(is_live)
         self.compareLiveButton.setChecked(is_live)
 
+    def launch_dialog(self):
+        self.dialog = RestoreDialog(self.client, self.snapshot)
```
tangkong, line 169: You should be sure to specify the parent of `RestoreDialog` is this widget, so when the parent widget is deleted any still-open `RestoreDialog` windows are also closed.

superscore/widgets/page/restore.py
```diff
@@ -110,12 +157,19 @@ def __init__(
...
+        self.dialog.restoreButton.clicked.connect(partial(self.tableView.set_live, True))
```
tangkong, line 170: In the past we've been weary of assigning partials to slots.  The reason is that the connection can persist even after the instance that holds the method is deleted.  ([see pcdsutils for our workaround and reasoning](https://github.com/pcdshub/pcdsutils/blob/359503d1c16f605bb81d78792bd32f5771666f84/pcdsutils/qt/callbacks.py#L17-L20))

We'll run into this less if the created `RestoreDialog` is properly parented (and thus can't persist after the restore page is closed), but it's still something to look out for.  This case may be fine (the child widget doesn't start any timers that might resolve after deletion, for example)

superscore/widgets/page/restore.py
shilorigins: `.close()` calls out that it doesn't actually delete the widget, so I wasn't sure if I had to do it or if that just meant it would get garbage collected.  I can switch to `.close()`, I'm pretty sure it works fine.

superscore/tests/ioc/ioc_factory.py
tangkong: Agreed 100%.  I think throughout the UI it's not unreasonable to expect the client to exist, and call its `.fill()` method or something similar.  (Maybe we add a `filled=True` optional argument to `Client.search()` or something as well)

I think in the test suite unless elsewise specified it's ok to have all the Entry's be filled.  I think if you ever use the Filestore backend you'll end up getting lazy `Entry`'s

superscore/widgets/page/restore.py
shilorigins: I tried opening the dialog and then closing the restore tab and the main window: in both cases the dialog also closed.  When I set the restore tab as the dialog's parent, the dialog opens attached to the tab rather than as a free-standing window.  I tried making the restore dialog an actual `QDialog` and had the same results.

Do you know a way to set the parent while keeping the widget a separate window?  And are we sure this is necessary?

shilorigins: I tried setting the dialog's parent in the constructor call instead of using `setParent()`, and that seems to work 🤷

shilorigins: Nevermind, it "worked" because the parent just wasn't being set

shilorigins: I think the issue was that `DesignerDisplay` doesn't inherit from any Qt class and we don't typically call the Qt superclass in our subclasses, so the parent arg in the dialog constructor wasn't getting used anywhere. I tried calling `QDialog.__init__` from the dialog's `__init__`: before `DesignerDisplay.__init__` causes the dialog to work but the parent is cleared, and after causes the parent to be correct but the dialog's contents are cleared.

I think I've sunk enough time into this for now, since the dialog already closes properly in my testing.  If we run into an issue with the dialog not closing properly, or if you have an approach to suggest, I'll revisit.

tangkong: Right, DesignerDisplay is more of a mixin that is meant to work with other QtWidget classes.  I've been using `DataWidget` as the other part of the mixin, but if you don't use that then you have to sort it yourself.  You should be calling the super constructor on most widgets you create though, if only to pass through other Qt specific arguments.  

Sorry for the confusion, QDialog does treat parenting a tad different than other widgets.  If it doesn't have issues being left dangling, then my fears are assuaged

shilorigins: The behaviour is nearly identical whether I use `QWidget` or `QDialog` as the other superclass, which is interesting. Calling `__init__` on both superclasses seems to clear some of the attributes, which is also interesting.  I'll leave it alone for now.

tangkong (approved):

superscore/widgets/page/restore.py
tangkong: I think it falls into the same trap, but we can leave it for now.  If we run into problems in the future we can loop back.

tangkong (approved):

PR #98, opened 2024-11-08
ENH: add Client.compare, along with dataclass diff calculation utilities

### LINKED ISSUE #9: Write `Entry` comparison method

tangkong:
## Description
- Adds `EntryDiff`, which tracks two `Entry`s, along with a list of `DiffItem` objects
  - attempts to build reasonable reprs for the above as well (WIP)
- implements `Client.compare` using the above

## Motivation and Context
closes #9 

## How Has This Been Tested?
Tests added.  

This PR, #9 

Example repr:
```
> client.compare( readback1, readback2 )
Diff: {
    Readback.uuid: (ecb42cdb-b703-4562-86e1-45bd67a2ab1a->8e380e15-5489-41db-a8a7-bc47a731f099)
    Readback.creation_time: (2024-05-10 16:49:34.574951+00:00->2024-05-10 16:49:34.574987+00:00)
    Readback.pv_name: (MY:PREFIX:mtr1.ACCL->MY:PREFIX:mtr1.VELO)
}
```

## Pre-merge checklist

### DIFF
--- a/superscore/client.py
+++ b/superscore/client.py
@@ -8,7 +8,7 @@ from uuid import UUID
 
 from superscore.backends import get_backend
 from superscore.backends.core import SearchTerm, SearchTermType, _Backend
-from superscore.compare import EntryDiff, walk_find_diff
+from superscore.compare import DiffItem, EntryDiff, walk_find_diff
 from superscore.control_layers import ControlLayer, EpicsData
 from superscore.control_layers.status import TaskStatus
 from superscore.errors import CommunicationError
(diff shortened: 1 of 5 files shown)


tangkong:
This is mostly in its final form, though I haven't fully qc'd myself.  I want to get eyes on this since it's sort of a strange construct.  If this looks ok I'll start thinking about how we might actually use these structures in the UI.  (Is making a table view using this information too inane?  Is it too much work to grab a diff for a specific entry?)

superscore/compare.py
```diff
@@ -0,0 +1,163 @@
...
+
+from superscore.model import Entry
+
+# An attribute access path chain leading to the item of interest
+# simple fields:        (object, "field_name")
+# dictionary values:    ("__dict__", "key_name")   ... not used?
+# list entry:           ("__list__", list_index: int)
+# enum value:           ("__enum__", Enum.member_object)
+# set entry:            ("__set__", "set_member")
+AttributePath = List[Tuple[Any, Any]]
```
ZLLentz, line 14: To help me understand the construction: are these examples all literal? As in, I may have the object literal or a string that's exactly `"__dict__"` or `"__list__"`, etc.?

superscore/compare.py
```diff
@@ -0,0 +1,163 @@
...
+
+def walk_find_diff(
+    orig_item: Union[Entry, Any],
+    new_item: Union[Entry, Any],
+    curr_path: Optional[AttributePath] = None,
+) -> Generator[DiffItem, None, None]:
+    if curr_path is None:
+        curr_path = []
+
+    if type(orig_item) is not type(new_item):
```
ZLLentz, line 79: Are there any cases where we want a more detailed diff where one item is a subclass of the other?

superscore/tests/test_compare.py
```diff
@@ -0,0 +1,177 @@
...
+from superscore.backends.core import SearchTerm
+from superscore.client import Client
+from superscore.compare import (AttributePath, DiffItem, EntryDiff,
+                                walk_find_diff)
+from superscore.model import (Collection, Entry, Parameter, Readback, Setpoint,
+                              Severity, Snapshot, Status)
+
+
+def simplify_path(path: AttributePath) -> AttributePath:
+    """swap objects for their type, for ease of testing"""
```
ZLLentz, line 16: To be clear, these are being used after running the function on the normal objects in order to allow you to more easily express the parameterized inputs/outputs?

ZLLentz (commented):
Some observations! I think some of this is pretty complicated but maybe it needs to be in order to handle the variety of data types.

superscore/tests/test_compare.py
tangkong: Yes that's exactly it.

superscore/client.py
```diff
@@ -186,9 +187,57 @@ def delete(self, entry: Entry) -> None:
...
+            diffs = DiffItem(original_value=entry_l, new_value=entry_r, path=[])
+            return EntryDiff(original_entry=entry_l, new_entry=entry_r, diffs=[diffs])
+
+        self.fill(entry_l)
+        self.fill(entry_r)
+        diffs = walk_find_diff(entry_l, entry_r)
+
+        return EntryDiff(original_entry=entry_l, new_entry=entry_r, diffs=list(diffs))
+
+    def fill(self, entry: Entry) -> None:
```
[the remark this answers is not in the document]
tangkong: In previous work, entries were filled on an as-needed basis, and often by calls in the GUI layer.  In those cases I think it makes sense to be less aggressive with the fill operation.

To my understanding this is the only method at the client level that fills `Entry`s, and the only method that fully fills the `Entry`.  It's worth re-assessing the other "filling" code.

superscore/compare.py
tangkong: I've mixed literal and placeholder representations here, I can try to clarify this in the docstring.  
* Position 0 in the tuple is always either the object being compared or literally the string as listed
* Position 1 either tells you how to access the object type (index in a list, key in a dict) or the member (for a set or enum)

superscore/compare.py
tangkong: As far as our model is concerned, I felt like this wasn't a relevant case.  All `Entry` children are direct children, and there are no grandchildren.  `Setpoint`, `Readback` and `Parameter` are similar, but we've made them distinct children of `Entry` at the same level.  We intend to treat these distinctly as well, so I felt like it made sense to not consider them similar wholesale.

I could be misguided here, though

superscore/compare.py
tangkong: an attempt to clarify this was made in https://github.com/pcdshub/superscore/pull/98/commits/2d4e19abcd7e6c27806d30b494022ad6f20a36a1

ZLLentz (approved):
I don't have any further thoughts on this- it seems well-motivated. Possibly we'll find some other cases later this doesn't cover that I can't think of now but we can handle that if we need to.

PR #107, opened 2024-11-19
MNT: move context menu functionality upstream into `RootTreeView`

### LINKED ISSUE #101: GUI: upstream context menu / click handling for Root Tree
## Current Behavior
The tree view in the main window has nice context menus and double-click-to-open-details functionality

All other tree views do not, since this was functionality added to the Window view only

## Expected Behavior
That cool stuff should exist for all the views 

## Context / environment
Raised during a dev discussion

N/A

## Suggested Solution
Create a common `RootTreeView` (analogous to `LivePVTableView` - `LivePVTableModel`) and upstream that functionality into the shared view.

This view will likely have to hold onto an open_page_slot (whether passed as an argument or set later on), to support those actions

tangkong:
## Description
- Adds common `RootTreeView` which holds the `RootTree` model
  - Upstreams standard settings and context menu support
- Fills the top-level entry whenever `RootTree` model is created, to ensure at least first level of children display

## Motivation and Context
Closes #101 

## How Has This Been Tested?
Interactively, and some tests have been added

Now you can double-click-to-open and right-click inside of other pages that have Tree Views
Now all pages with tree views will open with at least the first set of children populated.  

This PR

image

## Pre-merge checklist

### DIFF
--- a/superscore/widgets/views.py
+++ b/superscore/widgets/views.py
@@ -71,8 +71,15 @@ class EntryItem:
                 self._bridge_cache[id(data)] = bridge
                 self.bridge = bridge
 
-    def fill_uuids(self, client: Optional[Client] = None) -> None:
-        """Fill this item's data if it is a uuid, using ``client``"""
+    def fill_uuids(
+        self,
+        client: Optional[Client] = None,
+        fill_depth: int = 2
+    ) -> None:
+        """
+        Fill this item's data if it is a uuid, using ``client``.
+        By default fills to a depth of 2, to keep the tree view data loading lazy
+        """
         if client is None:
             return
 
@@ -82,7 +89,7 @@ class EntryItem:
 
         if isinstance(self._data, Nestable):
             if any(isinstance(child, UUID) for child in self._data.children):
-                client.fill(self._data, fill_depth=2)
+                client.fill(self._data, fill_depth=fill_depth)
 
             # re-construct child EntryItems if there is a mismatch or if any
             # hold UUIDs as _data
@@ -539,9 +546,6 @@ class RootTreeView(QtWidgets.QTreeView):
 
     @client.setter
     def client(self, client: Client):
-        self._set_client(client)
-
-    def _set_client(self, client: Client):
         if client is self._client:
             return
 
--- a/superscore/widgets/views.py
+++ b/superscore/widgets/views.py
@@ -288,6 +288,8 @@ class RootTree(QtCore.QAbstractItemModel):
         self.base_entry = base_entry
         self.root_item = build_tree(base_entry)
         self.client = client
+        # ensure at least the first set of children are filled
+        self.root_item.fill_uuids(self.client)
         self.headers = ['name', 'description']
 
     def refresh_tree(self) -> None:
--- a/superscore/widgets/views.py
+++ b/superscore/widgets/views.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import logging
 import time
 from enum import Enum, IntEnum, auto
+from functools import partial
 from typing import Any, ClassVar, Dict, Generator, List, Optional, Union
 from uuid import UUID
 from weakref import WeakValueDictionary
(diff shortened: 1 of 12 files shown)


superscore/widgets/window.py
```diff
@@ -55,13 +55,11 @@ def setup_ui(self) -> None:
...
-        self.tree_view.setModel(self.tree_model)
-        self.tree_view.setContextMenuPolicy(QtCore.Qt.CustomContextMenu)
-        self.tree_view.customContextMenuRequested.connect(self._tree_context_menu)
-        self.tree_view.setExpandsOnDoubleClick(False)
-        self.tree_view.doubleClicked.connect(self.open_index)
+        self.tree_view.client = self.client
+        self.tree_view.set_data(self.client.backend.root)
+        # override context menu and open_page_slot methods
+        self.tree_view.create_context_menu = self._window_context_menu
+        self.tree_view.open_page_slot = self.open_page
```
shilorigins, line 62: Do we need to override the context menu method?  `create_context_menu` and `_window_context_menu` are the same except for the restore page action, which I think should be present on embedded views as well.  So I think we could consolidate these into one `create_context_menu`

superscore/widgets/views.py
```diff
@@ -288,6 +289,8 @@ def __init__(
         self.base_entry = base_entry
         self.root_item = build_tree(base_entry)
         self.client = client
+        # ensure at least the first set of children are filled
+        self.root_item.fill_uuids(self.client)
```
shilorigins, line 293: I've just noticed that `EntryItem.fill_uuids` appears hard-coded to use `fill_depth=2`.  If it always fills only two levels, could we change the method name and docstring to make that clear?

superscore/widgets/views.py
```diff
@@ -497,6 +500,114 @@ def fetchMore(self, parent: QtCore.QModelIndex) -> None:
...
+    @client.setter
+    def client(self, client: Client):
+        self._set_client(client)
+
+    def _set_client(self, client: Client):
+        if client is self._client:
+            return
+
+        if not isinstance(client, Client):
+            raise ValueError("Provided client is not a superscore Client")
```
shilorigins, line 549: Out of curiosity, what's the benefit of introducing `_set_client` instead of putting the logic directly in `client.setter`'s body?

superscore/widgets/views.py
```diff
@@ -497,6 +500,114 @@ def fetchMore(self, parent: QtCore.QModelIndex) -> None:
...
+            logger.debug("Client not set, cannot initialize model")
+            return
+
+        if self.data is None:
+            logger.debug("data not set, cannot initialize model")
+            return
+
+        self._model = RootTree(base_entry=self.data, client=self.client)
+        self.setModel(self._model)
+        self._model.dataChanged.connect(self.data_updated)
```
shilorigins, line 574: Didn't know you could connect signals to signals. Neat!

superscore/widgets/window.py
tangkong: Currently we pass the `open_page_slot`'s through to the views already, but passing the `open_restore_page` through to all the child pages and views would be a bit of a bigger lift.  

I've been considering changing how we access these methods, as part of some further additions to the context menus.  I'll aim to tackle this in a later PR

superscore/widgets/views.py
tangkong: It was a pattern I copied from the table views, where some subclasses might do extra work after the `_set_client` work is finished.  But you're correct that here there's no real reason to do so.  I can consolidate these in the RootTreeView

superscore/widgets/window.py
shilorigins: I've had similar thoughts.  I recently used this block to find the main window and call the method I wanted directly:
```
window = self.parent()
while not isinstance(window, QtWidgets.QMainWindow):                                    
    window = window.parent()
```

We could also use a singleton.  Both of these require only ever wanting one `QMainWindow` of course.

Not sure what approaches you've considered, but wanted to throw out these couple thoughts.

tangkong: The singleton pattern is the main one I've been considering, with the caveat that we need to take care to avoid circular import shenanigans.  I prefer the singleton or some `get_main_window() -> QMainWindow` helper method to walking up a family tree that we may not have set properly, to be fair.

PR #108, opened 2024-11-21
ENH: add search term for retrieving Entries that are nested under an ancestor

### LINKED ISSUE #102: ENH: add ability to limit Entry searches bases on ancestor
Our search methods don't provide a way to limit results based on ancestor.  For example, I have a PV in one snapshot and I want to find the PV with the same name in a second snapshot.  There's no direct way to do this, because the search will return any `Setpoint`s with the same name in the entire backend, and there aren't other search terms that will properly limit the search.

This may work as a new `SearchTerm` operator, or as an arg in `Client.search()`.

shilorigins:
## Description

Add support for `ancestor` `SearchTerm` to `FilestoreBackend.search`
Add `FilestoreBackend._gather_progeny` utility method 

Closes #102 
## Motivation and Context

With an `Entry` in one `Nestable`, it is impossible to retrieve the corresponding `Entry` in another `Nestable` using the existing `SearchTerms`.  The attributes of the known `Entry` do not reliably correspond to attributes of the desired `Entry`, other than the `.pv_name` or `.title`, so it is necessary to be able to limit searches to `Entry`s that are nested under a known `Nestable`.

A concrete use case is comparing snapshots.  For each PV in the first snapshot, the corresponding PV must be found in the second snapshot using only the PV string.  This becomes possible using the new search term.

The current implementation uses a utility method to gather all `Entry`s nested under an ancestor `Entry`.  This enables caching the result using `functools.cache`, which lets us avoid:
1. Traversing the ancestor's DAG for each entry we check
2. Iterating over the search terms in advance in order to gather all progeny before checking the search conditions

Calling `cache` directly rather than as a decorator limits the caching to each `FilestoreBackend.search` call, so changes to the data between searches are reflected properly. I believe each cache should be garbage collected after each search, but I haven't been able to verify this yet.

## How Has This Been Tested?

A unit test in `tests/test_backend` and an integration test in `tests/test_client`.

## Pre-merge checklist

### DIFF
--- a/superscore/backends/filestore.py
+++ b/superscore/backends/filestore.py
@@ -264,7 +264,7 @@ class FilestoreBackend(_Backend):
             if db.get(entry.uuid):
                 raise BackendError("Entry already exists, try updating the entry "
                                    "instead of saving it")
-            db[entry.uuid] = entry
+            self.flatten_and_cache(entry)
             self._root.entries.append(entry)
 
     def update_entry(self, entry: Entry) -> None:
--- a/superscore/backends/filestore.py
+++ b/superscore/backends/filestore.py
@@ -8,14 +8,15 @@ import logging
 import os
 import shutil
 from dataclasses import fields, replace
-from typing import Any, Dict, Generator, Optional, Union
+from functools import cache
+from typing import Any, Container, Dict, Generator, Optional, Union
 from uuid import UUID, uuid4
 
 from apischema import deserialize, serialize
 
 from superscore.backends.core import SearchTermType, _Backend
 from superscore.errors import BackendError
-from superscore.model import Entry, Root
+from superscore.model import Entry, Nestable, Root
 from superscore.utils import build_abs_path
 
 logger = logging.getLogger(__name__)
@@ -285,6 +286,7 @@ class FilestoreBackend(_Backend):
         Keys are attributes on `Entry` subclasses, or special keywords.
         Values can be a single value or a tuple of values depending on operator.
         """
+        reachable = cache(self._gather_reachable)
         with self._load_and_store_context() as db:
             for entry in db.values():
                 conditions = []
@@ -292,6 +294,8 @@ class FilestoreBackend(_Backend):
                     # TODO: search for child pvs?
                     if attr == "entry_type":
                         conditions.append(isinstance(entry, target))
+                    elif attr == "ancestor":
+                        conditions.append(entry.uuid in reachable(target))
                     else:
                         try:
                             # check entry attribute by name
@@ -301,6 +305,22 @@ class FilestoreBackend(_Backend):
                             conditions.append(False)
                 if all(conditions):
                     yield entry
+
+    def _gather_reachable(self, ancestor: UUID) -> Container[UUID]:
+        """
+        Finds all entries accessible from ancestor, including ancestor, and returns
+        their UUIDs. This makes it easy to check if one entry is hierarchically under another.
+        """
+        reachable = set()
+        q = [ancestor]
+        while len(q) > 0:
+            cur = q.pop()
+            if not isinstance(cur, Entry):
+                cur = self._entry_cache[cur]
+            reachable.add(cur.uuid)
+            if isinstance(cur, Nestable):
+                q.extend(cur.children)
+        return reachable
 
     @contextlib.contextmanager
     def _load_and_store_context(self) -> Generator[Dict[UUID, Any], None, None]:
--- a/superscore/backends/filestore.py
+++ b/superscore/backends/filestore.py
@@ -295,6 +295,7 @@ class FilestoreBackend(_Backend):
                     if attr == "entry_type":
                         conditions.append(isinstance(entry, target))
                     elif attr == "ancestor":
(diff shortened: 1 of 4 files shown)


superscore/backends/filestore.py
```diff
@@ -302,6 +306,22 @@ def search(self, *search_terms: SearchTermType) -> Generator[Entry, None, None]:
                 if all(conditions):
                     yield entry
 
+    def _gather_reachable(self, ancestor: UUID) -> Container[UUID]:
```
tangkong, line 309: The ancestor here could also work as `Union[Entry, UUID]` right?  Is there a reason we would require this to start from a UUID?

superscore/tests/test_client.py
```diff
@@ -211,8 +211,24 @@ def test_fill_depth(fill_depth: int):
 
 
 @pytest.mark.parametrize("filestore_backend", [("linac_with_comparison_snapshot",)], indirect=True)
-def test_parametrized_filestore(sample_client: Client):
```
tangkong, line 214: We didn't like this test anymore?  I guess it was a pretty minimal smoke test

superscore/tests/test_backend.py
```diff
@@ -202,3 +202,16 @@ def test_update_entry(backends: _Backend):
...
+def test_gather_reachable(filestore_backend: _Backend):
+    # top-level snapshot
+    reachable = filestore_backend._gather_reachable(UUID("06282731-33ea-4270-ba14-098872e627dc"))
+    assert len(reachable) == 32
+    assert UUID("927ef6cb-e45f-4175-aa5f-6c6eec1f3ae4") in reachable
+
+    # direct parent snapshot
+    reachable = filestore_backend._gather_reachable(UUID("2f709b4b-79da-4a8b-8693-eed2c389cb3a"))
+    assert len(reachable) == 3
+    assert UUID("927ef6cb-e45f-4175-aa5f-6c6eec1f3ae4") in reachable
```
tangkong, line 217: I went about verifying that the `functools.cache` scope behaved as you claimed by augmenting this test case a bit.  Things worked as advertised, and `_gather_reachable` seems to work well when handed an `Entry` as well.
 
```suggestion
    assert UUID("927ef6cb-e45f-4175-aa5f-6c6eec1f3ae4") in reachable
    
    entry = filestore_backend.get_entry(UUID("2f709b4b-79da-4a8b-8693-eed2c389cb3a"))
    assert len(filestore_backend._gather_reachable(entry)) == 3
    entry.children = []
    assert len(filestore_backend._gather_reachable(entry)) == 1
    filestore_backend.update_entry(entry)
    reachable = filestore_backend._gather_reachable(UUID("2f709b4b-79da-4a8b-8693-eed2c389cb3a"))
    assert len(reachable) == 1
```

tangkong (dismissed):
Looks good to me 👍   I left suggestions for a slightly expanded test and a more permissive type hint, but I think the core of this is hard to disagree with.  I imagine this gets refactored a bit with the Visitor concept

superscore/tests/test_client.py
shilorigins: The new test implicitly covers the old test, but we could keep the old one if we want the more explicit name if an error occurs.

tangkong: nah, I just tend to never look at consolidating old tests, it's better as it is 👍

superscore/backends/filestore.py
shilorigins: The parameters to a cached function have to hashable, that's the only thing. But if passing an `Entry` worked in the testing, then I'm not sure.  I might want to expand the test even more to be sure.

tangkong: Ah that's a very good callout.  If that's the case, I imagine it's enough to force/require arguments to be UUIDs in `.update_entry()`, since that's the only place that actually invokes `functools.cache`.  

This method itself makes sure that only hashable uuids make it into the `reachable` set

shilorigins: Passing an `Entry` to `FilestoreBackend._gather_reachable` works because the method isn't cached by default.  The method is cached within the body of `FilestoreBackend.search`, so inside `.search` it has to be called with UUIDs only, but outside of that it can be called with UUIDs or `Entry`s.

I'm not sure what you mean about `.update_entry`, did you mean `.search`?

superscore/tests/test_backend.py
shilorigins: I don't think a cache is ever actually created in this test.  `_gather_reachable` isn't cached on its own, `FilestoreBackend.search` caches it within its method body, so we need `.search` calls to test caching behaviour.  I've moved this suggestion into `test_client::test_search_entries_by_ancestor`.

superscore/backends/filestore.py
tangkong: Yes I think we're agreeing with each other here, just with different words 👍 

github diff view gets me again, yes I mean `.search()` 🤦

superscore/tests/test_backend.py
tangkong: You are correct 👍

tangkong (approved):

Issue #106, opened 2024-11-15
[Prospective] MNT: consolidate different Entry traversal use cases in Visitor


shilorigins:
We re-implement the logic of traversing an `Entry` graph multiple times in our codebase, performing a different action at each node or tracking different metadata.  These use cases could potentially be streamlined by implementing an `EntryVisitor` class containing the common logic, with subclasses performing the different specific functions.

Possible visitor subclasses:
* `FillUUIDVisitor`
* `SearchVisitor` (for `FilestoreBackend` / `TestBackend`)
* `SnapVisitor`

tangkong:
Or possibly a visitor + post-visit hook structure?  I realize that's the normal way you handle graph/tree traversal

PR #110, opened 2024-11-26
Devagr/entry visitor

shilorigins:
## Description

Implement `EntryVisitor` abstract class, with corresponding `Entry.accept` methods
Implement `SearchVisitor` as the first concrete visitor
Add `ancestor` search option to `SearchVisitor`

Relates to #106 
## Motivation and Context

Using a visitor pattern will let us decouple `Entry` DAG traversal from specific components, such as the `Client` or concrete `_Backend` subclasses. This should make it easier to do things like collect and search entries, fill UUIDs, and save snapshots without having to re-implement traversal logic in multiple places.
## How Has This Been Tested?

Existing search tests cover the new `SearchVisitor`.
## Pre-merge checklist

- [ ] Code works interactively
- [ ] Code follows the [style guide](https://pcdshub.github.io/style.html)
- [ ] Code contains descriptive docstrings, including context and API
- [ ] Test suite passes locally
- [ ] Test suite passes on GitHub Actions
- [ ] Ran ``docs/pre-release-notes.sh`` and created a pre-release documentation page
- [ ] Pre-release docs include context, functional descriptions, and contributors as appropriate

### DIFF
--- a/superscore/backends/test.py
+++ b/superscore/backends/test.py
@@ -9,6 +9,7 @@ from superscore.backends.core import SearchTermType, _Backend
 from superscore.errors import (BackendError, EntryExistsError,
                                EntryNotFoundError)
 from superscore.model import Entry, Nestable, Root
+from superscore.visitor import SearchVisitor
 
 
 class TestBackend(_Backend):
@@ -75,18 +76,7 @@ class TestBackend(_Backend):
         return self._root
 
     def search(self, *search_terms: SearchTermType):
-        for entry in self._entry_cache.values():
-            conditions = []
-            for attr, op, target in search_terms:
-                # TODO: search for child pvs?
-                if attr == "entry_type":
-                    conditions.append(isinstance(entry, target))
-                else:
-                    try:
-                        # check entry attribute by name
-                        value = getattr(entry, attr)
-                        conditions.append(self.compare(op, value, target))
-                    except AttributeError:
-                        conditions.append(False)
-            if all(conditions):
-                yield entry
+        visitor = SearchVisitor(self, *search_terms)
+        root = self.root
+        visitor.visit(root)
+        yield from visitor.matches
--- a/superscore/visitor.py
+++ b/superscore/visitor.py
@@ -1,0 +1,49 @@
+""""""
+from collections.abc import Union
+from uuid import UUID
+
+from superscore.backends import _Backend
+from superscore.model import (Collection, Entry, Parameter, Readback, Root,
+                              Setpoint, Snapshot)
+
+
+class EntryVisitor:
+    """"""
+    def __init__(self, backend: _Backend):
+        self.backend = backend
+
+    def visit(self, entry: Union[Entry, Root, UUID]) -> None:
+        if isinstance(entry, UUID):
+            entry = self.backend.get_entry(entry)
+        entry.accept(self)
+        return
+
+    def visitParameter(self, parameter: Parameter) -> None:
+        raise NotImplementedError
+
+    def visitSetpoint(self, setpoint: Setpoint) -> None:
+        raise NotImplementedError
+
+    def visitReadback(self, readback: Readback) -> None:
+        raise NotImplementedError
+
+    def visitCollection(self, collection: Collection) -> None:
+        raise NotImplementedError
+
+    def visitSnapshot(self, snapshot: Snapshot) -> None:
+        raise NotImplementedError
+
+    def visitRoot(self, root: Root) -> None:
+        raise NotImplementedError
+
+
+class FillUUIDVisitor(EntryVisitor):
+    pass
+
+
+class SnapVisitor(EntryVisitor):
+    pass
+
+
+class SearchVisitor(EntryVisitor):
+    pass
--- a/superscore/visitor.py
+++ b/superscore/visitor.py
@@ -1,10 +1,21 @@
 """"""
-from collections.abc import Union
+import re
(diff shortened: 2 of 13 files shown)


superscore/visitor.py
```diff
@@ -0,0 +1,133 @@
...
+    def __init__(self, backend):
+        self.backend = backend
+
+    def visit(self, entry: Union[Entry, Root, UUID]) -> None:
+        if isinstance(entry, UUID):
+            entry = self.backend.get_entry(entry)
+        entry.accept(self)
+        return
+
+    def visitParameter(self, parameter: Parameter) -> None:
```
tangkong, line 23: We can possibly use [`functools.singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch) here to consolidate these methods.  (as I familiarize myself more with this pattern I realize that's exactly what we should do)

superscore/backends/test.py
```diff
@@ -75,18 +76,7 @@ def root(self) -> Root:
...
-                        # check entry attribute by name
-                        value = getattr(entry, attr)
-                        conditions.append(self.compare(op, value, target))
-                    except AttributeError:
-                        conditions.append(False)
-            if all(conditions):
-                yield entry
+        visitor = SearchVisitor(self, *search_terms)
+        root = self.root
+        visitor.visit(root)
```
tangkong, line 81: I'm seeing this elsewhere as starting with an "accept" call, such as `root.acept(visitor)`.  These are functionally equivalent right?

superscore/visitor.py
tangkong: I'm actually confused by the presence of these two types of visit method, really all `.visit()` is doing here is calling `Entry.accept()`, which ultimately calls a `.visit*()` method?  (In a sort of triple dispatch back and forth).  Maybe all the `.visit()`'s are meant to be interchangeable?

superscore/visitor.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class SearchVisitor(EntryVisitor):
+    def __init__(self, backend, *search_terms: Iterable[SearchTerm]):
+        super().__init__(backend)
+        self.search_terms = search_terms
+        self.path = []
+        self.matches = []
+
+    def _check_match(self, entry: Union[Entry, Root]) -> bool:
```
tangkong, line 57:
```suggestion
    def _check_match(self, entry: Union[Entry, Root]) -> None:
```
Because it confused me as I was reading this.  

This would be the place to expose any entry that matches, and I think it might be worth bubbling that up and making `visit` yield as well.  I can imagine other visitors gathering PVs or specific entries etc, but not every node we visit will yield a result (Entries that don't match).  So in the end maybe only `EntryVisitor.visit` returns a generator and whatever other worker methods return values as necessary?

In untested skeleton code I'm imagining here (consolidating all the visit_<ENTRYTYPE> methods):
```python
    def visit(self, entry: Union[Entry, Root, UUID]) -> None:
        if isinstance(entry, UUID):
            entry = self.backend.get_entry(entry)
        self.path.append(entry)

        # Perform work
        result = self._check_match(entry)
        if result is not None:
            yield result

        # visiting entry's children handled by Entry.accept() calls, previously in super().visit(entry)
        # but we get to visit via an `Entry.accept()` call, so why do we call accept again here?

        self.path.pop()
```
Excuse my general confusion here

tangkong (commented):
I think the presence and differentiation between `.visit()` and `.visitEntryType()` methods is throwing me off here, left a few rather confused comments.  Perhaps consolidating everything with single-dispatch will make things simpler / smoother

superscore/backends/test.py
shilorigins: You're right, looks like I made up `.visit` without realizing.  It does give us a benefit though: we can use `.visit` for tracking state during traversal.  For example, `SearchVisitor` adds the `Entry` to the path and pops it off the path in `.visit` because `.visit` doesn't return until all of the reachable entries have been traversed, whereas `.visit*` return before the reachable entries are traversed.  Without `.visit`, we might have to move the traversal logic into `.visit*`, so we can track the state we need.

I'll try out removing `.visit` and seeing what changes need to be made.

superscore/visitor.py
shilorigins: It looks like I made up `.visit` without realizing.  There's a little more discussion in the below thread.

superscore/backends/test.py
tangkong: For reference, a sequence diagram and class diagram from the source of all my knowledge: https://en.wikipedia.org/wiki/Visitor_pattern#Structure

superscore/visitor.py
shilorigins: This is basically how I was thinking to use a generator.  Per the above thread, if we remove `.visit` then the generator might become `.accept`, but the idea is basically the same.

RE consolidating `.visit*`: In `SearchVisitor` all of the `.visit*` methods are identical because we searches treat the different `Entry` subclasses the same: we check if they match the search terms.  Other visitors might treat the subclasses differently, like `SnapVisitor` only saving data for `Parameter`s.  So in general we can't consolidate the `.visit*` methods, but in some specific visitors they'll be identical.

tangkong: I agree that the visitors want to differentiate between `Entry` subclasses, I was arguing more for doing that via single dispatch and overloading the `.visit()` method.

In that case we won't have to make the `Entry.accept` methods pick the correct `.visit` method, the `Visitor` will get the correct one at runtime

Issue #48, opened 2024-07-01
GUI: Build Entry comparison page (Snapshots only?)


tangkong:

tangkong:
This looks like it relies on #9 , @shilorigins did you already start work on the client-level comparison function?  If not I can take a crack at it

shilorigins:
I haven't.  But this issue seems like it overlaps with the restore page functionality of comparing snapshots. That one is focused on PV values, is this one focused on metadata or something other than the contained PVs?

tangkong:
At the very least, I think that this comparison should happen at the client level, rather than in the UI.  Ideally the UI calls this client function, and picks the data that it cares about.  

If we sequester this functionality inside the UI, we won't be able to access it from other apps that only use the client.

shilorigins:
Sure, and I can write the snapshot portion of the client method since I'll use it for the restore page.  But I was also thinking about how this comparison page will handle snapshots, because of that overlap between the pages.

tangkong:
I want to take a look at this for comparing arbitrary Entry's, I think a complete treatment of this probably looks pretty different from the existing snapshot restore page

PR #112, opened 2024-12-04
GUI: add Diff-view page

tangkong:
## Description
- Adds a Diff Page that subclasses the existing common views to add highlighting based on the calculated Diff
- Adds a QtSingleton class that allows singletons to hold signals.
  - Long story short, the normal `__new__` singleton creation method doesn't quite work.  While the singletons are created correctly, the bound `QtCore.Signal`'s lose their connections every time an instance is returned
  - Devan theorizes that this has to do with `__init__` doing something to the class variable signals.  I am too tired to investigate further

This description is so minimal for the amount of time I spent spinning on this 💀 

This is likely still riddled with bugs, but I want to get eyes on it before it bloats into a small planet of a diff.  I'm happy to revisit this in follow-up PRs, but I feel like there are other more pressing tasks to take on.  (And I'm tired of thinking about diffs)

Outstanding tasks:
- add comparison context menus to main window tree view (this overrides the common views, so I left it for now)
- Touchups on diff highlighting.  (I'm bad at logic-ing things)

## Motivation and Context
#48 in part

## How Has This Been Tested?
I added a smoke test.  I should have probably added more

This PR.  I'll go through for a docstring pass shortly

Context menus have been added to all non-main-window views 
image

Highlighting helps, but could be more thorough / accurate
image

HIghlighting for entries with PV-entries
image

## Pre-merge checklist

### DIFF
--- a/superscore/compare.py
+++ b/superscore/compare.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, fields, is_dataclass
+from enum import Enum, auto
 from typing import Any, Generator, Iterable, List, Optional, Tuple, Union
 
 from superscore.model import Entry
@@ -12,6 +13,12 @@ from superscore.model import Entry
 # enum value:           ("__enum__", enum_member: Enum)
 # set entry:            ("__set__", set_member: str)
 AttributePath = List[Tuple[Any, Any]]
+
+
+class DiffType(Enum):
+    DELETED = auto()
+    MODIFIED = auto()
+    ADDED = auto()
 
 
 @dataclass
@@ -48,6 +55,17 @@ class DiffItem:
         repr_str += f": ({orig_val_str}->{new_val_str})"
 
         return repr_str
+
+    @property
+    def type(self) -> DiffType:
+        if (self.original_value is not None) and (self.new_value is None):
+            return DiffType.DELETED
+        elif (self.original_value is None) and (self.new_value is not None):
+            return DiffType.ADDED
+        elif (self.original_value is not None) and (self.new_value is not None):
+            return DiffType.MODIFIED
+        else:
+            raise ValueError("DiffItem original and new values are both None")
 
 
 @dataclass
--- a/superscore/tests/test_page.py
+++ b/superscore/tests/test_page.py
@@ -11,6 +11,7 @@ from superscore.control_layers._base_shim import EpicsData
 from superscore.model import (Collection, Parameter, Readback, Setpoint,
                               Snapshot)
 from superscore.widgets.page.collection_builder import CollectionBuilderPage
+from superscore.widgets.page.diff import DiffPage
 from superscore.widgets.page.entry import (BaseParameterPage, CollectionPage,
                                            ParameterPage, ReadbackPage,
                                            SetpointPage, SnapshotPage)
@@ -90,6 +91,35 @@ def restore_page(qtbot: QtBot, sample_client: Client, simple_snapshot: Snapshot)
     page.close()
 
 
+@pytest.fixture
+def diff_page(qtbot: QtBot, sample_client: Client):
+    l_snapshot = Snapshot(
+        title="l_snap",
+        description="l desc",
+        children=[
+            Setpoint(pv_name="MY:SET", data=1),
+            Readback(pv_name="MY:RBV", data=1),
+        ]
+    )
+    r_snapshot = Snapshot(
+        title="r_snap",
+        description="r desc",
+        children=[
+            Setpoint(pv_name="MY:SET", data=2),
+            Readback(pv_name="MY:RBV", data=1),
+            Readback(pv_name="MY:RBV2", data=2),
+        ]
+    )
+    page = DiffPage(
+        client=sample_client,
+        l_entry=l_snapshot,
+        r_entry=r_snapshot,
+    )
+    qtbot.addWidget(page)
+    yield page
+    page.close()
+
+
 @pytest.mark.parametrize(
     'page',
     [
@@ -101,6 +131,7 @@ def restore_page(qtbot: QtBot, sample_client: Client, simple_snapshot: Snapshot)
         "search_page",
         "collection_builder_page",
         "restore_page",
+        "diff_page",
     ]
 )
 def test_page_smoke(page: str, request: pytest.FixtureRequest):
--- a/superscore/widgets/core.py
+++ b/superscore/widgets/core.py
@@ -8,7 +8,7 @@ from typing import ClassVar, List
 from weakref import WeakValueDictionary
 
 from pcdsutils.qt.designer_display import DesignerDisplay
-from qtpy import QtWidgets
+from qtpy import QtCore, QtWidgets
 
 from superscore.qt_helpers import QDataclassBridge, QDataclassList
 from superscore.type_hints import AnyDataclass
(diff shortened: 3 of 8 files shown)


tangkong:
I'd like to request reviews here, focused mostly on the design/organization of the code.  If you have the patience to QA the diff highlighting, I welcome that as well, but I do intend to circle back to that in the near future.

superscore/widgets/page/diff.py
```diff
@@ -0,0 +1,421 @@
...
+                                      RootTreeView)
+
+logger = logging.getLogger(__name__)
+
+
+DIFF_COLOR_MAP = {
+    DiffType.DELETED: QtGui.QColor(255, 0, 0, alpha=100),
+    DiffType.MODIFIED: QtGui.QColor(255, 255, 0, alpha=100),
+    DiffType.ADDED: QtGui.QColor(0, 255, 0, alpha=100),
+    None: None,
```
ZLLentz, line 28: What is this None case used for?

superscore/widgets/page/diff.py
```diff
@@ -0,0 +1,421 @@
...
+
+
+class BiDict(dict):
+    """
+    Bi-directional mapping dictionary.
+    Every key-value pair can only appear once in each dict (forward and inverse)
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.inverse = {}
```
ZLLentz, line 39: General design question: does having a separate `inverse` dict have some advantages over putting the inverse mapping into the same dictionary as the forward mapping?

E.g. this could have also been implemented as
dict[key] = value
(internally, also do dict[value] = key)

superscore/widgets/page/diff.py
```diff
@@ -0,0 +1,421 @@
...
+
+    def __setitem__(self, key, value):
+        if key in self:
+            self.inverse[self[key]].remove(key)
```
ZLLentz, line 43: This implies that all objects in the `inverse` dictionary have a method called `remove`. That seems incorrect. Is this line meant to be something else?

superscore/widgets/views.py
```diff
@@ -25,13 +26,103 @@
...
+        f'&Open Detailed {type(entry).__name__} page'
+    )
+    # WeakPartialMethodSlot may not be needed, menus are transient
+    # TODO: refactor to not require passing open_page_slot, maybe with window
+    # singleton eventually
+    open_action.triggered.connect(partial(open_page_slot, entry))
+
+
+# Entries for comparison
+class DiffDispatcher(QtCore.QObject, metaclass=QtSingleton):
```
ZLLentz, line 52: It's not obvious to me what the value is in having this be a singleton. Will this be created and re-used in multiple contexts? What's the consequence of these contexts not referring to the same singleton object? Would it have been possible to refer to and reference the same object again?

ZLLentz (commented):
I don't think I recommended any actions in my comments, it's all questions, so maybe I'm just a confused person. The code is clean so it was easy to focus on the small things.

superscore/compare.py
```diff
@@ -49,6 +56,17 @@ def __repr__(self) -> str:
 
         return repr_str
 
+    @property
+    def type(self) -> DiffType:
+        if (self.original_value is not None) and (self.new_value is None):
```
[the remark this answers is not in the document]
tangkong: This is a good point, there are some fields in our dataclasses that allow `None` (optional fields).  Though I don't think this changes how the `DiffItem` interprets values going from something -> `None`, this would still be a deletion as far as the diff types go.  (or an addition for `None`-> something)

superscore/compare.py
```diff
@@ -49,6 +56,17 @@ def __repr__(self) -> str:
...
+            return DiffType.DELETED
+        elif (self.original_value is None) and (self.new_value is not None):
+            return DiffType.ADDED
+        elif (self.original_value is not None) and (self.new_value is not None):
+            return DiffType.MODIFIED
+        else:
+            raise ValueError("DiffItem original and new values are both None")
```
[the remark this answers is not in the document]
tangkong: Currently this is true; `DiffItems` are only created when modifications are detected (or at least that's my intention).  I imagine it's most concise to have only modifications show up, otherwise we'll have a `DiffItem` for every single field in every dataclass.

superscore/tests/test_page.py
```diff
@@ -90,6 +91,35 @@ def restore_page(qtbot: QtBot, sample_client: Client, simple_snapshot: Snapshot)
...
+            Readback(pv_name="MY:RBV", data=1),
+        ]
+    )
+    r_snapshot = Snapshot(
+        title="r_snap",
+        description="r desc",
+        children=[
+            Setpoint(pv_name="MY:SET", data=2),
+            Readback(pv_name="MY:RBV", data=1),
+            Readback(pv_name="MY:RBV2", data=2),
```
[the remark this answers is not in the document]
tangkong: This is certainly not a sufficient test case.  I definitely need to build out the tests, I was hoping to skate by this time and circle back to it.

superscore/widgets/page/diff.py
tangkong: This just made it simpler for me to inspect the dict through the normal repr, it looks like a normal dict but I can access either the key or the value.  Your implementation works similarly well, I imagine.

superscore/widgets/page/diff.py
tangkong: The `None` case is the "No modification" case, where the `.data(index, role=Qt.BackgroundColor)` call just returns None and doesn't change the background color.  It's sort of my fallback case, I should add a comment here explaining that

superscore/widgets/page/diff.py
tangkong: ... erm yes I don't know what happened there.  this should be `del self.inverse[self[key]]`

superscore/widgets/views.py
tangkong: Yes, this is meant to allow us to signal between widgets without navigating through a parent-child chain.  

In this case, the `views` module holds onto the left/right entries used for comparison, and uses that information to create the context menus for the common table/tree views.  By having this dispatcher be a singleton, we can subscribe to that signal from the top level Window and open the diff page when the `DiffDispatcher` emits.  We grab this singleton `DiffDispatcher` (really primarily for holding a singleton signal) in the `Window` widget and also each of the common views.

Without this, we'd have to fall back to the open_page_slot pattern where we pass a callable through to every page through to the common views.  This has already become onerous, and I've considered making the Main window a singleton so that we can access its members easily from any child page (without walking up a questionably connected parent chain).  

> Would it have been possible to refer to and reference the same object again?
This is what I was trying to achieve with the singleton construction.  Do you mean achieving this not with some forced singleton, but with a single object that we make/get with some factory function helper?  I thought that this approach was a bit more bulletproof, guarding against unintentional module-level access.  (Or maybe I've misinterpreted this comment)

ZLLentz: I understand now! Thanks

superscore/widgets/core.py
```diff
@@ -396,3 +396,30 @@ def on_del_clicked(self, **kwargs) -> None:
...
+
+    """
+    def __init__(cls, name, bases, dict):
+        super().__init__(name, bases, dict)
+        cls._instance = None
+
+    def __call__(cls, *args, **kwargs):
+        if cls._instance is None:
+            cls._instance = super().__call__(*args, **kwargs)
+        return cls._instance
```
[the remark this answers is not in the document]
tangkong: Yes, for the most part.  The other way singletons may be created in python classes is with the `__new__` dunder method, caching the first instance of the class in a similar way but with a different dunder method.  

This effectively returned a singleton with each instantiation request, but also cleared signal connections along with it.  I don't entirely understand why this happens, but I do know the metaclass approach with `__call__` works.

A rabbit hole I'll jump down at some other point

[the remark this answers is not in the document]
ZLLentz: I bet it has something to do with the precise way pyqt bridges the gap between python and the C objects for the purpose of qt signals...

tangkong:
@ZLLentz Any other thoughts here?  Bugging you because Devan is out this week and I want to build on top of this

ZLLentz:
I have nothing else- my questions have all been fully answered and the diff looks nice

ZLLentz (approved):

PR #115, opened 2024-12-11
REF: create a window singleton for access throughout superscore

tangkong:
## Description
- Makes `Window` a `QtSingleton`, and adds an access function for use throughout superscore ui widgets
  - Uses this singleton to access methods that were passed around manually in the past as optional arguments or set attributes.  In particular
    - Open Page Slot: removed as an optional argument on all page widgets, only retrieved by the Window singleton.  (since this argument has no meaning outside the context of the full application)
    - client: Added a backup access path via `Window` singleton, but also retained the init argument.  This permits testing of widgets individually
- These methods are gathered in a `WindowLinker` mixin class for convenience and de-duplication 

I ended up treating these two differently, and am open to differing opinions here.  Perhaps I should have left the existing open_page_slot optional arguments and treated it like I did the Client

## Motivation and Context
We found that we were passing around information held by the window throughout our application.  We also find ourselves wanting to open pages from other sub-pages, not just from the main window.  

If we ever want to change the information passed from page to page, we'd end up having to change the signature of all those widgets. It's probably best to avoid this behavior

## How Has This Been Tested?
By ensuring the tests still pass.  The required arguments of the existing widgets haven't changed, in theory (as borne out by the tests)

This PR

## Pre-merge checklist

### DIFF
--- a/superscore/tests/test_page.py
+++ b/superscore/tests/test_page.py
@@ -1,6 +1,6 @@
 """Largely smoke tests for various pages"""
 
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock, PropertyMock, patch
 
 import pytest
 from pytestqt.qtbot import QtBot
@@ -219,10 +219,11 @@ def test_open_page_slot(
     page_fixture: str,
     request: pytest.FixtureRequest,
 ):
-    page: BaseParameterPage = request.getfixturevalue(page_fixture)
-    page.open_page_slot = MagicMock()
-    page.open_rbv_button.clicked.emit()
-    assert page.open_page_slot.called
+    with patch("superscore.widgets.page.entry.BaseParameterPage.open_page_slot",
+               new_callable=PropertyMock):
+        page: BaseParameterPage = request.getfixturevalue(page_fixture)
+        page.open_rbv_button.clicked.emit()
+        assert page.open_page_slot.called
 
 
 @pytest.mark.parametrize(
--- a/superscore/widgets/__init__.py
+++ b/superscore/widgets/__init__.py
@@ -20,10 +20,11 @@ ICON_MAP = _get_icon_map()
 
 def get_window():
     """
-    Return the window singleton if it already exists, to allow creating widgets
-    without creating Window
+    Return the window singleton if it already exists, to allow other widgets to
+    access its members.
     Must not be called in the code path that results from Window.__init__.
-    A good (safe) rule of thumb is never reached from any widget's __init__.
+    A good (safe) rule of thumb is to make sure this function cannot be reached
+    from any widget's __init__ method.
     Hides import in __init__ to avoid circular imports.
     """
     from .window import Window
--- a/superscore/widgets/__init__.py
+++ b/superscore/widgets/__init__.py
@@ -16,3 +16,17 @@ def _get_icon_map():
 
 
 ICON_MAP = _get_icon_map()
+
+
+def get_window():
+    """
+    Return the window singleton if it already exists, to allow creating widgets
+    without creating Window
+    Must not be called in the code path that results from Window.__init__.
+    A good (safe) rule of thumb is never reached from any widget's __init__.
+    Hides import in __init__ to avoid circular imports.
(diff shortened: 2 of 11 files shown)


tangkong:
perhaps there's also a world where we pull these methods up into a common core class, but I dislike how that obfuscates these very frequently used methods.  (Is the duplication worth it?  maybe not)

superscore/tests/test_page.py
```diff
@@ -219,10 +219,11 @@ def test_open_page_slot(
     page_fixture: str,
     request: pytest.FixtureRequest,
 ):
-    page: BaseParameterPage = request.getfixturevalue(page_fixture)
-    page.open_page_slot = MagicMock()
-    page.open_rbv_button.clicked.emit()
-    assert page.open_page_slot.called
+    with patch("superscore.widgets.page.entry.BaseParameterPage.open_page_slot",
```
ZLLentz, line 222: Test design question: when would you choose `unittest.mock.patch` over the pytest `monkeypatch` fixture and vice versa?

superscore/widgets/__init__.py
```diff
@@ -16,3 +16,17 @@ def _get_icon_map():
...
 
 ICON_MAP = _get_icon_map()
+
+
+def get_window():
+    """
+    Return the window singleton if it already exists, to allow creating widgets
+    without creating Window
+    Must not be called in the code path that results from Window.__init__.
+    A good (safe) rule of thumb is never reached from any widget's __init__.
```
ZLLentz, line 26: Nit: the grammar here is a bit scrambled

superscore/widgets/page/collection_builder.py
```diff
@@ -48,14 +48,12 @@ def __init__(
...
         data: Optional[Collection] = None,
-        open_page_slot: Optional[OpenPageSlot] = None,
         **kwargs
     ):
         if data is None:
             data = Collection()
         super().__init__(*args, data=data, **kwargs)
-        self.client = client
-        self.open_page_slot = open_page_slot
+        self._client = client
```
ZLLentz, line 56: I don't understand why this needed to be changed to the _private variable. It looks to me like we would have been just fine using the property setter from `WindowLinker`.

superscore/widgets/core.py
```diff
@@ -423,3 +425,43 @@ def __call__(cls, *args, **kwargs):
...
+
+    @client.setter
+    def client(self, client: Client):
+        if not isinstance(client, Client):
+            raise TypeError(f"Cannot set a {type(client)} as a client")
+
+        if client is self._client:
+            return
+
+        self._client = client
```
ZLLentz, line 461: I notice that this allows different widgets to use different clients. Is this intended?

ZLLentz (commented):
I'm on board conceptually and the implementation looks good. I have some questions and specific things that confused me.

superscore/tests/test_page.py
tangkong: I think there's not much of a difference most of the time.  In this case I wanted to mock a property, and the first example I found was through unittest.mock.  I've been using `MagicMock` and finding it very convenient, but it is some deep mystical magic

I should probably be better about sticking to one, but we have access to both.

superscore/widgets/__init__.py
tangkong: yes, it appears my brain stopped brain-ing...

superscore/widgets/core.py
tangkong: Within the context of the superscore app, this isn't something we expect to happen.  We should only have one client for one superscore instance and all its child pages.  I'm not particularly worried about a new client getting slipped into the superscore app, but this does provide that possibility

Personally, specifying this per widget lets me test widgets individually, free of needing to create a `Window`.

superscore/widgets/page/collection_builder.py
tangkong: You are correct, I forgot to remove this.

ZLLentz (approved):
I'll put an approval check here

PR #118, opened 2025-02-21
REF: Make CLI subcommand structure more lazy, defer heavy imports

### LINKED ISSUE #87: BUG: error messages while importing subcommand modules are suppressed
## Current Behavior

If an error occurs while loading a subcommand module (e.g. `bin/ui.py`), the specific error message is suppressed and the program exits with a generic `invalid choice` message.
## Expected Behavior

Errors that occur while loading subcommand modules should be shown to the user, at least when that subcommand is used.
## Context / environment

During development, the obfuscation of error messages forces me to run through a debugger to reproduce the error message or to carefully inspect the file changes.

1. In the global namespace of `bin/ui.py`, add a line `raise Exception("Reason for the error")` 
2. Run `superscore ui`
3. Note the output: 'ui' is not a valid choice, but no reason is given

tangkong:
## Description
Moves heavier imports to *_main.py files, allowing subcommand argparsers to be collected independently

## Motivation and Context
Some python libraries have heavier imports (qt I'm looking at you), and we shouldn't need to import them if we're simply looking for cli subcommands or help-text.  Though a bit more boilerplate, separating the "functional" imports into another file allows us to defer that loading until we actually invoke the subcommand

I'll fix #87 here too at some point

## How Has This Been Tested?
interactively, and pytest runs

This PR. 

## Pre-merge checklist


superscore/bin/main.py
```diff
@@ -26,13 +26,25 @@ def _try_import(module):
...
+    Also builds up the the argparse description whenever a subcommand is added.
+    """
     global DESCRIPTION
     result = {}
     unavailable = []
 
     for module in sorted(MODULES):
         try:
-            mod = _try_import(module)
+            mod = _try_import(module + "_parser")
```
ngAlexander9, line 47: I think this might be causing an issue with the help command.
image

tangkong: Thanks for catching this!

tangkong:
I think this is also ready to go, if people don't have any objections

shilorigins (approved):
I confirmed all of the commands work and this speeds up the usage messages.  I'm fine with these changes

PR #125, opened 2025-02-28
BUG: refresh main window tree when new collection is created

### LINKED ISSUE #111: BUG: creating new collection doesn't add to tree view
## Current Behavior
Creating a new collection doesn't refresh the window-level tree view, and there's no way to refresh it manually

## Expected Behavior
Adding a collection should be reflected promptly

## Context / environment
Was poking around

1. Open superscore demo
2. File > New > Collection
3. Add pvs and save
4. Notice tree does not update
5. Go to search page and search for collection, it should appear

## Suggested Solution
Trigger tree refresh with collection save?
add refresh button on window for tree view?

tangkong:
## Description
Refresh the main window tree view whenever a Collection is added through the Collection Builder Page

## Motivation and Context
Closes #111 

This takes advantage of the Window singleton pattern added some time ago.  Using this prevents us from having to pass specific methods/signals through to widgets.

## How Has This Been Tested?
Interactively, a simple test has been added.

This PR 

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/125-bug_refresh_window.rst
+++ b/docs/source/upcoming_release_notes/125-bug_refresh_window.rst
@@ -1,0 +1,22 @@
+125 bug_refresh_window
+######################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- Refreshes window whenever a new collection is added from the `CollectionBuilderPage`
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tangkong
--- a/superscore/tests/test_window.py
+++ b/superscore/tests/test_window.py
@@ -5,6 +5,8 @@ from pytestqt.qtbot import QtBot
 from superscore.backends.filestore import FilestoreBackend
 from superscore.client import Client
 from superscore.tests.conftest import setup_test_stack
+from superscore.widgets.page.collection_builder import CollectionBuilderPage
+from superscore.widgets.views import EntryItem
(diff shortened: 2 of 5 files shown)


shilorigins:
Works as described.  My only concern is with the call to `window.tree_view.set_data(self.client.backend.root)`.  Is there no way to update the tree view without rebuilding the entire model?

tangkong:
(whoops left this dangling for a bit)

This is a good question, and maybe it's something we revisit.

I think it is necessary due to the way that the tree grabs data.  When the tree data is first set, we grab `client.backend.root`, which us up-to-date at the time.  This root is constructed when we make the call to the property, so it isn't necessarily updated when made edits to the backend (saving/deleting/modifying).  An additional call to root is necessary to see these changes.

When we're sure that the base entry has all the necessary children (we're simply filling uuid data, or we're building a collection), we can simply tell the view that the model data is changing via `QAbstractItemModel.layoutAboutToBeChanged` as [here](https://github.com/pcdshub/superscore/blob/2be6a866a478a1971e2af59e9cc0609c9771166a/superscore/widgets/views.py#L393-L396)


Looking forward, if we start to collect global views like this, it might be worth grouping them and possibly setting up a better signaling system.  Right now it's up to us to remember to refresh the main tree view

tangkong:
Any reviews here? Seems like #128 wants to make use of this, so it  makes sense to try to get this in?

shilorigins (approved):
Whoops forgot to officially approve

tangkong:
Let the rebasing begin!

PR #128, opened 2025-03-05
ENH: add button to take snapshots on Collection / Snapshot detail page

### LINKED ISSUE #50: GUI: Build "save a Snapshot" page (tracks progress of save)
Page to go from collection to snapshot.

Should:
- display values as they are collected
- provide fields for adding metadata to the newly collected snapshot
- ...

### LINKED ISSUE #126: BUG: seemingly random error in filestore::fill_uuids
After checking out a new branch and running `superscore ui`, I got this error:
```
(.env) lcls-dev3:~/src/superscore/superscore$ superscore ui
Traceback (most recent call last):
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/.env/bin/superscore", line 8, in <module>
    sys.exit(main())
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/superscore/bin/main.py", line 91, in main
    func(**kwargs)
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/superscore/bin/ui.py", line 26, in main
    main_window = Window(client=client)
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/superscore/widgets/core.py", line 426, in __call__
    cls._instance = super().__call__(*args, **kwargs)
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/superscore/widgets/window.py", line 51, in __init__
    self.setup_ui()
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/superscore/widgets/window.py", line 63, in setup_ui
    self.tree_view.set_data(self.client.backend.root)
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/superscore/backends/filestore.py", line 249, in root
    with self._load_and_store_context():
  File "/afs/slac/g/lcls/package/anaconda/envs/python3_rhel7_env/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/superscore/backends/filestore.py", line 340, in _load_and_store_context
    self.store()
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/superscore/backends/filestore.py", line 206, in store
    self._root = self.reconstruct_root()
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/superscore/backends/filestore.py", line 146, in reconstruct_root
    new_children.append(self.fill_uuids(new_child))
  File "/afs/slac.stanford.edu/u/cd/devagr/src/superscore/superscore/backends/filestore.py", line 188, in fill_uuids
    setattr(entry, field.name)
TypeError: setattr expected 3 arguments, got 2
```

After confirming that the test suite ran and passed, I couldn't reproduce the error.  Just wanted to log the incident, since the `setattr` call is in fact missing a parameter.

shilorigins:
## Description

* fix bug where `Client._build_snapshot` calls `_Backend.get`, which doesn't exist
* fix bug where `FilestoreBackend.fill_uuids` calls `setattr` with only two args
  * closes #126 
* strip `AugmentedValue` metadata when setting `EpicsData.data` in `_AoicaShim`
  * this was causing a bug where returning `EpicsData.data` from a view's `.data` method caused the cells to show as empty
* add button on `nestable_page.ui` (used for showing details of `Collection`s and `Snapshot`s) that takes a new snapshot for the associated `Collection`
  * closes #50 
* include dialog for setting metadata for new snapshot
## Motivation and Context

Taking new snapshots is a core feature of the app.  The button is placed on the detailed nestables page so that it's available for `Collection`s and `Snapshot`s.

The dialog uses a `NameDescTagsWidget`, which requires attaching a `Snapshot` to leverage the existing functionality.  Since `Client._build_snapshot` creates a new `Snapshot` and then builds its structure using the origin `Collection` for reference, passing the newly made `Snapshot` through to `Client._build_snapshot` saves some arg checks and redundant variable assignments.

`Client.snap` first fetches values for all of the PVs, then creates and builds the new `Snapshot`, so the app currently hangs while values are being collected rather than opening the snapshot table and filling in the cells as values arrive.  Since I would have to refactor `Client.snap` / `._build_snapshot` to achieve the latter behaviour, I've omitted it from this Pr.

## How Has This Been Tested?

New test (in `test_window.py` since it requires inspecting the tab widget and tree view).

This PR

## Pre-merge checklist

### DIFF
--- a/superscore/backends/filestore.py
+++ b/superscore/backends/filestore.py
@@ -185,7 +185,7 @@ class FilestoreBackend(_Backend):
                 new_ref = self._entry_cache.get(field_data)
                 if new_ref:
                     new_ref = self.fill_uuids(new_ref)
-                    setattr(entry, field.name)
+                    setattr(entry, field.name, new_ref)
 
         return entry
 
--- a/superscore/tests/test_window.py
+++ b/superscore/tests/test_window.py
@@ -80,3 +80,29 @@ def test_add_collection_refresh(qtbot: QtBot, test_client: Client):
     new_top_level_entries = new_entry_item.childCount()
 
     assert new_top_level_entries > orig_top_level_entries
+
+
+@setup_test_stack(sources=["db/filestore.json"], backend_type=FilestoreBackend)
+def test_take_snapshot(qtbot, test_client):
+    window = Window(client=test_client)
+    qtbot.addWidget(window)
+    collection = tuple(test_client.search(("uuid", "eq", UUID("a9f289d4-3421-4107-8e7f-2fe0daab77a5"))))[0]
+    snapshot = tuple(test_client.search(("uuid", "eq", UUID("ffd668d3-57d9-404e-8366-0778af7aee61"))))[0]
+
+    collection_page = window.open_page(collection)
+    new_snapshot = collection_page.take_snapshot()
+    collection_page.children()[-1].done(1)
+    search_result = tuple(test_client.search(("uuid", 'eq', new_snapshot.uuid)))
+    assert new_snapshot == search_result[0]
+
+    snapshot_page = window.open_page(snapshot)
+    result = snapshot_page.take_snapshot()
+    assert result is None
+    snapshot_page.close()
+
+    snapshot.origin_collection = collection.uuid
+    snapshot_page = window.open_page(snapshot)
+    new_snapshot = snapshot_page.take_snapshot()
+    snapshot_page.children()[-1].done(1)
+    search_result = tuple(test_client.search(("uuid", 'eq', new_snapshot.uuid)))
+    assert new_snapshot == search_result[0]
--- a/superscore/widgets/core.py
+++ b/superscore/widgets/core.py
@@ -466,6 +466,10 @@ class WindowLinker:
         if window is not None:
             return window.open_page
 
+    def get_window(self):
+        """Return the singleton Window instance"""
+        return get_window()
+
     def refresh_window(self):
         """Refresh window ui elements"""
         # tree view
--- a/superscore/widgets/page/entry.py
+++ b/superscore/widgets/page/entry.py
@@ -4,12 +4,14 @@ Widgets for visualizing and editing core model dataclasses
 import logging
 from copy import deepcopy
 from typing import Optional, Union
+from uuid import UUID
 
 import qtawesome as qta
 from qtpy import QtWidgets
 from qtpy.QtGui import QCloseEvent
 
 from superscore.control_layers._base_shim import EpicsData
+from superscore.errors import EntryNotFoundError
 from superscore.model import (Collection, Nestable, Parameter, Readback,
                               Setpoint, Severity, Snapshot, Status)
 from superscore.type_hints import AnyEpicsType
@@ -35,6 +37,7 @@ class NestablePage(Display, DataWidget, WindowLinker):
     sub_pv_table_view: LivePVTableView
 
     save_button: QtWidgets.QPushButton
+    snapshot_button: QtWidgets.QPushButton
 
     data: Nestable
 
@@ -67,6 +70,8 @@ class NestablePage(Display, DataWidget, WindowLinker):
         self.sub_coll_table_view.data_updated.connect(self.track_changes)
 
         self.save_button.clicked.connect(self.save)
+        self.snapshot_button.clicked.connect(self.take_snapshot)
+        self.snapshot_button.setText("Take new Snapshot")
 
         self.set_editable(self.editable)
 
@@ -100,6 +105,30 @@ class NestablePage(Display, DataWidget, WindowLinker):
         logger.debug(f"Stopping polling threads for {type(self.data)}")
         self.sub_pv_table_view._model.stop_polling(wait_time=5000)
         return super().closeEvent(a0)
+
+    def take_snapshot(self) -> None:
+        # TODO: if data dirty, abort
+        entry = self.data
+        if isinstance(entry, Snapshot):
+            origin = entry.origin_collection
+            if origin is None:
+                logging.error("The original collection is unknown: cannot save snapshot")
+                return
+            elif isinstance(origin, UUID):
+                try:
+                    entry = self.client.backend.get_entry(origin)
+                except EntryNotFoundError:
+                    logging.error("The original collection does not exist: cannot save snapshot")
+                    return
+            else:
+                entry = origin
+        window = self.get_window()
+
+        snapshot = self.client.snap(entry)
+        self.client.save(snapshot)
+        self.refresh_window()
+        window.open_restore_page(snapshot=snapshot)
+        return
 
 
 class CollectionPage(NestablePage):
--- a/superscore/widgets/page/entry.py
+++ b/superscore/widgets/page/entry.py
@@ -3,8 +3,8 @@ Widgets for visualizing and editing core model dataclasses
 """
 import logging
 from copy import deepcopy
+from functools import partial
 from typing import Optional, Union
-from uuid import UUID
 
 import qtawesome as qta
 from qtpy import QtWidgets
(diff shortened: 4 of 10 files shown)


superscore/backends/filestore.py
```diff
@@ -185,7 +185,7 @@ def fill_uuids(self, entry: Entry) -> Entry:
                 new_ref = self._entry_cache.get(field_data)
                 if new_ref:
                     new_ref = self.fill_uuids(new_ref)
-                    setattr(entry, field.name)
+                    setattr(entry, field.name, new_ref)
```
tangkong, line 188: wow we just never got here huh...

superscore/widgets/page/entry.py
```diff
@@ -101,6 +107,65 @@ def closeEvent(self, a0: QCloseEvent) -> None:
...
+            dialog = self.prompt_for_metadata(dest_snapshot)
+            dialog.accepted.connect(partial(self.client.snap, dest_snapshot.origin_collection, dest=dest_snapshot))
+            dialog.accepted.connect(partial(self.client.save, dest_snapshot))
+            dialog.accepted.connect(self.refresh_window)
+
+            window = self.get_window()
+            dialog.accepted.connect(partial(window.open_restore_page, snapshot=dest_snapshot))
+            return dest_snapshot
+
+    def find_origin_collection(self, entry: Union[Collection, Snapshot]) -> Collection:
```
tangkong, line 131: I wonder if this should be a `Client` method, it seems like it'd be useful and something we might end up implementing again in a different context.

superscore/widgets/page/entry.py
```diff
@@ -101,6 +107,65 @@ def closeEvent(self, a0: QCloseEvent) -> None:
...
+        """
+        Return the Collection instance associated with an entry.  The entry can
+        be a Collection or Snapshot.
+
+        Raises
+        ------
+        ValueError
```
tangkong, line 138: Is raising appropriate here?  We list `origin_collection` as `Optional` in the model definitions so _technically_ having not having an origin is valid.

I imagined a world where someone might create a snapshot without a collection, at least via the python interface.  In the GUI I could see how that might not happen, I suppose

superscore/widgets/core.py
```diff
@@ -466,9 +466,13 @@ def open_page_slot(self) -> Optional[OpenPageSlot]:
...
 
+    def get_window(self):
+        """Return the singleton Window instance"""
+        return get_window()
+
     def refresh_window(self):
         """Refresh window ui elements"""
         # tree view
-        window = get_window()
+        window = self.get_window()
```
tangkong, line 476: is there a difference here between `self.get_window()` and `get_window()`?  I ask only because we change this reference but not any of the others.

superscore/tests/test_page.py
tangkong, line 1: No tests for the widget itself? :(

tangkong (commented):
Sorry for the delay in review!  This looks pretty good to me, just a few nits/questions.  I'll take it for a quick spin first thing tomorrow to play QA monkey

superscore/widgets/core.py
shilorigins: I wanted to get the window instance from within `NestablePage`.  It seemed like `WindowLinker` was intended as the intermediary between subclasses and globals like the window singleton and the client, plus `NestablePage` already subclassed `WindowLinker`, so I could get the instance that route instead of importing it directly.

In this file, it doesn't make any difference.  I can change all the references to either option to be consistent.

superscore/widgets/page/entry.py
shilorigins: I agree

superscore/widgets/page/entry.py
shilorigins: The origin isn't mandatory for the data model, but it is theoretically mandatory for certain actions.  As you say, a user can construct a snapshot without an origin, and they can restore such a snapshot.  The question here is whether a user should be able to save a new snapshot based on the structure of a manually constructed snapshot.  I intuitively thought no, manual snapshots shouldn't be able to spawn programmatic snapshots, but this is open to discussion.

Taking the above to be true, I chose to handle those edge cases via exceptions rather than special return values.  If the `find_origin_collection` is supposed to return the filled entry associated with a `Nestable`, what do we do when the origin isn't set or when the origin is set but isn't in the backend?  We could return `None`, but the two cases are slightly different.  I didn't think of a suitable second reserved return value, so I went with exceptions.  This also lets us simply propagate `_Backend.get_entry`'s implicit `EntryNotFoundError` instead of catching it simply to return a reserved value.

superscore/tests/test_page.py
tangkong: I usually use `setText` type methods to set data, then emit the `clicked` signals to trigger slots.  I think `QtBot` can also send mouse clicks, which should be equivalent

superscore/widgets/core.py
tangkong: Yeah I'm all for making the main window accessible from its children, just wondering if there was something squirrely going on that made plain `get_window` not work.  I'm all for consistency but we're nitpicking at that point

superscore/widgets/page/entry.py
tangkong: I sort of naively thought returning `None` when there was no origin collection made sense.  

I worry about the potentially arbitrary line between "theoretically mandatory" and "technically mandatory", and feel like it'd make our lives easier if they were always kept in alignment.  

I think it's worth some more discussion.  I thought the case where we'd want to create a snapshot based off another orphan snapshot wasn't too far from reality.  Perhaps if we allow actions like this, we should just create the origin collection to match.

🤔

tangkong: Interestingly enough, the demo database snapshots don't have origin_collections set, so even though the "take snapshot" button exists on the snapshot NestablePage, it fails throwing an error

shilorigins: I added an origin to the demo snapshots, so taking snapshots off of those works now.

tangkong: This technically satisfies my comment, I just don't want to lose the thread here on whether or not snapshots should have mandatory origins.  I'll slap an issue out to record the thought

superscore/widgets/page/entry.py
```diff
@@ -101,6 +106,41 @@ def closeEvent(self, a0: QCloseEvent) -> None:
         self.sub_pv_table_view._model.stop_polling(wait_time=5000)
         return super().closeEvent(a0)
 
+    def take_snapshot(self) -> Snapshot:
```
tangkong, line 109:
```suggestion
    def take_snapshot(self) -> Optional[Snapshot]:
```

If the snapshot doesn't have an origin collection this just returns None.

The choice to throw Exceptions in `Client.find_origin_collection` and return `None` here still seems a little at odds, I feel like we should choose one approach.  Or maybe I'm still hung up on the fact that we throw exceptions when an optional field is `None`

superscore/widgets/page/entry.py
```diff
@@ -101,6 +106,41 @@ def closeEvent(self, a0: QCloseEvent) -> None:
...
+        layout = QtWidgets.QVBoxLayout()
+        layout.addWidget(NameDescTagsWidget(data=dest))
+        buttonBox = QtWidgets.QDialogButtonBox(
+            QtWidgets.QDialogButtonBox.Save | QtWidgets.QDialogButtonBox.Cancel
+        )
+        layout.addWidget(buttonBox)
+        buttonBox.accepted.connect(metadata_dialog.accept)
+        buttonBox.rejected.connect(metadata_dialog.reject)
+        metadata_dialog.setLayout(layout)
+        metadata_dialog.open()
```
tangkong, line 141: Here we're opening the dialog before we've connected the accepted slots.  If some user is spamming an accept hotkey, is it possible that we accept and close the dialog before the slots are connected?  (in some super fast spam or super slow computer situation)

I'd probably rather return the dialog and open it once connections are established

superscore/tests/test_window.py
```diff
@@ -80,3 +80,29 @@ def test_add_collection_refresh(qtbot: QtBot, test_client: Client):
...
 
     assert new_top_level_entries > orig_top_level_entries
+
+
+@setup_test_stack(sources=["db/filestore.json"], backend_type=FilestoreBackend)
+def test_take_snapshot(qtbot, test_client):
+    window = Window(client=test_client)
+    qtbot.addWidget(window)
+    collection = tuple(test_client.search(("uuid", "eq", UUID("a9f289d4-3421-4107-8e7f-2fe0daab77a5"))))[0]
+    snapshot = tuple(test_client.search(("uuid", "eq", UUID("ffd668d3-57d9-404e-8366-0778af7aee61"))))[0]
```
tangkong, line 90:
```suggestion
    snapshot = tuple(test_client.search(("uuid", "eq", UUID("ffd668d3-57d9-404e-8366-0778af7aee61"))))
```

These are generators right?

tangkong (commented):
Sorry more comments on a second readthrough

superscore/widgets/page/entry.py
shilorigins: I don't see much of a difference between returning `None` and raising an error in the lower method. In both cases, the caller just has to implement a check, either an `is None` or an `except`.  `find_origin_collection` can propagate `EntryNotFoundError`s from `backend.get_entry` anyway, so using a `ValueError` just made the handling cleaner.

Since I attach `take_snapshot` to a button, I don't know how to handle raised exceptions, which is why I return `None`.

superscore/widgets/page/entry.py
shilorigins: Good point

superscore/widgets/page/entry.py
tangkong: Yeah I'll put down the issue of none vs exceptions, but I do think the type hint should be updated.

Sorry for being annoying about it 😓

superscore/tests/test_window.py
tangkong: sorry my suggestion was nonsensical at first.  I've adjusted it to just use `next`

shilorigins: I was just looking at that haha.  I think this is more pythonic.

superscore/widgets/page/entry.py
shilorigins: Right, I'll still update the type hint.

My reasoning for using an error is that in my mind `find_origin_collection` is supposed to return the `Collection` entry associated with a nestable, and if it can't find such an entry, then that's an error.  Sure `Snapshot.origin_collection` is optional, but if it's `None` then `find_origin_collection` can't do its job because there's no entry to fetch.

But like I said, I don't actually see much pragmatic difference, so I don't mind changing it if it's bothersome.

tangkong: I think your rationale makes sense, the only difference in my mind being that `find_origin_collection` returning `None` still seems to me like it's doing the job, there's just no origin to return.

No need to change it.  If it's a problem in the future we'll find out and fix it then.

superscore/tests/test_window.py
tangkong: Unpacking the generator here is causing a problem in a subsequent call to Client.fill?...

shilorigins: That's the implication, but I haven't yet isolated that happening

shilorigins: So, the backend's entry cache goes from 10 entries down to 4 entries if I call `next` until reaching `StopIteration`.  Maybe the exception is preventing the cache from being restored?

tangkong: Ah I think I see the suspect code.  The backend's `.search` method yields from within a cache-modification decorator, so maybe when we don't exit that decorator immediately we end up throwing the cache into a weird intermediate state.

https://github.com/pcdshub/superscore/blob/dede6666a6f5761f8de5ef8913785827b19380c3/superscore/backends/filestore.py#L296-L315

I can't say I have a solution since I haven't reproduced the problem myself, but I'd bet it's in there somewhere

shilorigins: The error persists without the cache.  However, calling `next` on both generators in the test a second time (resulting in `StopIteration`) fixes the error.  So I think the issue is ultimately nested `FilestoreBackend._load_and_store_context`s when the generator isn't exhausted.

shilorigins: In that case I think I'll leave the code as-is.  `tuple(generator)` seems more intuitive than a hardcoded sequence of `next` calls.

tangkong: Sounds good, this is a separate bug to squash for sure.

tangkong (approved):
This looks good (and has since the beginning despite my nitpicks)

PR #129, opened 2025-03-07
ENH: Add meta PVs to top-level entries in demo dataset and implement table to present them.

### LINKED ISSUE #120: TST: add meta PVs and values to test data

ngAlexander9:
## Description

- Add `.vscode` directory to gitignore
- Add a `Parameter` and `Readback` representing some beam energy PV, and it to the top level collection and snapshot, respectively.
- Fit tab group into existing layout, allowing for switching between live 'content' PVs and meta PVs (label text tentative)
- Implement `LiveMetaPVTableView`, which is derived from `LivePVTableView`, but looks at data from `self.data.meta_pvs`

image

## Motivation and Context
Adding meta pvs to the demo dataset will help demonstrate that functionality.
Resolves #120 

This PR

## Pre-merge checklist

- [ ] Code works interactively
- [ ] Code contains descriptive docstrings, including context and API
- [ ] Test suite passes locally
- [ ] Test suite passes on GitHub Actions
- [ ] Ran ``docs/pre-release-notes.sh`` and created a pre-release documentation page
- [ ] Pre-release docs include context, functional descriptions, and contributors as appropriate

### DIFF
--- a/.github/workflows/standard.yml
+++ b/.github/workflows/standard.yml
@@ -1,7 +1,6 @@
 name: PCDS Standard Testing
 
 on:
-  push:
   pull_request:
   release:
     types:
--- a/.gitignore
+++ b/.gitignore
@@ -114,3 +114,5 @@ Thumbs.db
 
 # setuptools-scm _version.py should not be tracked:
 /superscore/_version.py
+
+.vscode/*
--- a/superscore/tests/conftest_data.py
+++ b/superscore/tests/conftest_data.py
@@ -302,6 +302,14 @@ def linac_data() -> Root:
         ]
     )
 
+    all_col_beam_e_meta_pv = Parameter(
+        uuid="e4aa8dc8-dbb0-4331-999a-89053758d489",
+        pv_name="BEAME",
+        description="A beam energy PV",
+    )
+
+    all_col.meta_pvs.append(all_col_beam_e_meta_pv)
+
     lasr_gunb_value1 = Setpoint(
         uuid="927ef6cb-e45f-4175-aa5f-6c6eec1f3ae4",
         pv_name=lasr_gunb_pv1.pv_name,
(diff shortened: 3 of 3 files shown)


ngAlexander9:
Additional effort needed for unit tests + fixing implementation of remove_row method

superscore/widgets/page/entry.py
```diff
@@ -65,6 +67,14 @@ def setup_ui(self):
         self.sub_pv_table_view.set_data(self.data)
         self.sub_pv_table_view.data_updated.connect(self.track_changes)
 
+        self.sub_meta_pv_table_view.client = self.client
```
ngAlexander9, line 70: Hiding the columns only seems to work after setting client and data, otherwise I would have put these in the init of LiveMetaPVTableView

superscore/widgets/views.py
```diff
@@ -1529,6 +1529,69 @@ def closeEvent(self, a0: QtGui.QCloseEvent) -> None:
...
+                    raise EntryNotFoundError(f"{child} not found in backend, "
+                                             "cannot fill with real data")
+
+                if not isinstance(child, Nestable) and isinstance(child, Entry):
+                    self.sub_entries.append(child)
+
+        elif isinstance(self.data, (Parameter, Setpoint, Readback)):
+            self.sub_entries = [self.data]
+
+    def remove_row(self, index: QtCore.QModelIndex) -> None:
```
ngAlexander9, line 1581: This isn't a fully working implementation. Currently, I think the method does remove the row from the model, but the table is not redrawn

tangkong: You have to signal to the view that it should be re-drawn if you're going to modify the underlying model.  There's a variety of signals to send, in this case `dataChanged` might be sufficient

https://doc.qt.io/qt-6/model-view-programming.html#making-the-model-editable

PR #131, opened 2025-03-12
BUG: fix incorrect arguments to WindowLinker.open_page_slot

tangkong:
## Description
- Fixes incorrect arguments to open_page_slot, which I missed when transitioning to use `WindowLinker`
- Also adjusts ResultFilterProxyModel to make a call to the `WindowLinker` method instead of calling `get_window ` itself.

## Motivation and Context
I got embarrassed by this bug during the standup and figured I should fix it

## How Has This Been Tested?
Interactively.  I'm going to skip writing a test for this one because triggering a context menu programmatically would involve too much hard coded position wrangling.  Maybe I revisit it later...

This PR

## Pre-merge checklist

### DIFF
--- a/superscore/widgets/views.py
+++ b/superscore/widgets/views.py
@@ -41,9 +41,7 @@ def add_open_page_to_menu(
     open_action = menu.addAction(
         f'&Open Detailed {type(entry).__name__} page'
     )
-    # WeakPartialMethodSlot may not be needed, menus are transient
-    # TODO: refactor to not require passing open_page_slot, maybe with window
-    # singleton eventually
+    # WeakPartialMethodSlot may not be needed, menus are transients
     open_action.triggered.connect(partial(window.open_page, entry))
 
 
@@ -694,8 +692,9 @@ class RootTreeView(QtWidgets.QTreeView, WindowLinker):
         """
         menu = QtWidgets.QMenu(self)
 
+        # checking if window exists to attach open_page_slot to
         if self.context_menu_options[MenuOption.OPEN_PAGE] and self.open_page_slot:
-            MENU_OPT_ADDER_MAP[MenuOption.OPEN_PAGE](menu, entry, self.open_page_slot)
+            MENU_OPT_ADDER_MAP[MenuOption.OPEN_PAGE](menu, entry)
 
         menu.addSeparator()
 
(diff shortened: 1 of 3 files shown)


tangkong:
This has been dangling for a while, any reviewers?

shilorigins:
Generally, I'm a little confused about the motivation behind having a mix-in that abstracts calls to a singleton.  Why is subclassing `WindowLinker` and then calling `self.open_page_slot` better than calling `get_window` and then `Window.open_page`?

tangkong:
> Generally, I'm a little confused about the motivation behind having a mix-in that abstracts calls to a singleton. Why is subclassing `WindowLinker` and then calling `self.open_page_slot` better than calling `get_window` and then `Window.open_page`?

It was mostly to avoid breaking existing pages, which used to store the callable in that attribute.  Moving forward calling get_window is fine, but we can't import `Window` for type hints due to circular imports iirc

The original effort to pull these up into a common mixin was for de-duplication.  Whether or not it's the best approach is up for debate.

superscore/widgets/views.py
```diff
@@ -1447,7 +1449,7 @@ def create_context_menu(self, entry: Entry) -> QtWidgets.QMenu:
         """
         menu = QtWidgets.QMenu(self)
         if self.context_menu_options[MenuOption.OPEN_PAGE] and self.open_page_slot:
```
shilorigins, line 1451:
```suggestion
        if self.context_menu_options[MenuOption.OPEN_PAGE]:
```
Same as above thread

shilorigins:
> > Generally, I'm a little confused about the motivation behind having a mix-in that abstracts calls to a singleton. Why is subclassing `WindowLinker` and then calling `self.open_page_slot` better than calling `get_window` and then `Window.open_page`?
> 
> It was mostly to avoid breaking existing pages, which used to store the callable in that attribute. Moving forward calling get_window is fine, but we can't import `Window` for type hints due to circular imports iirc
> 
> The original effort to pull these up into a common mixin was for de-duplication. Whether or not it's the best approach is up for debate.

I'm not sure that `WindowLinker` is reducing duplication much, like the views that subclass it end up overwriting the client setter anyway. It seems like the client getter is seeing use, but I was thinking it might make sense to add a client singleton for similar reasons as `Window`.  Overall I'd say I prefer the clarity from manipulating the singleton directly over the type hint.

tangkong:
I appreciate the critical eye to the Window singleton structure 👍 

The views override the client method but all the other `Entry` pages use it as is, for example.  

The client singleton is an interesting thought, currently it's effectively a singleton since we only access it through the `Window`.  I think we're more likely to see multiple clients (read multiple data sources/backend) in one superscore process than we are to see multiple main windows.

PR #152, opened 2026-04-10
ENH: import / export functionality

tangkong:
## Description
- Adds import / export functionality as a menu option
  - importing simply opens the requested page in a new page, for the user to save manually
  - These new pages get new uuids, the serialized uuids get ignored (this is unclear but there's no easy way to omit it without re-rolling a serialization framework or mangling the datastructure)
- Requires Snapshots to have existing origin collections during a client.save
- Adds (de)serialization helpers to `superscore.model`
  - we wrap these in a `Root` object to let the tagged union subclass detection ffom apischema work.
- Refactors validation methods/helpers to return a `ValidationResult` dataclass rather than a simple boolean, allowing reasons to be passed through the application
  - These are bool-like, so they should be useable in the same contexts as before
- adds "open origin" button to snapshot page
- Adds logic to entry pages to check validation status as a condition for enabling the save button, complete with semi helpful tooltip

TODO: (deferred to later effort)
- adjust pages to be dirty after import

## Motivation and Context
Requested by Usman as the final feature for Superscore before beta release.  For opening the door to external-to-superscore workflows
image
image

A note of caution, currently there is no way to link a snapshot to a specific collection.  If you want to upload a valid snapshot you'll have to know what uuid to reference as the origin collection.  This is trivially done by exporting an existing valid snapshot and modifying it.
There is also no verification that the parent collection matches the child snapshot, but they asked for arbitrary data and there's only so much I want to patch around the back door🤷 

### General musings
Every time I look at this project I uncover a slew of database validation conditions that may or may not belong in the client. 

Because the db has no logic, we keep this in application code, but I should probably be clearer about what logic lives where to avoid fragmentation

Frankly, this request makes more work for us since we have to now ingest and validate json blobs given to us by random users.  However It's a good escape hatch and we'll need to eventually be measured about validation anyway.

## How Has This Been Tested?
tests have been added

This PR

## Pre-merge checklist

### DIFF
--- a/superscore/client.py
+++ b/superscore/client.py
@@ -21,6 +21,7 @@ from superscore.model import (Collection, Entry, Nestable, Parameter, Readback,
 from superscore.templates import (TemplateMode, fill_template_collection,
                                   find_placeholders)
 from superscore.utils import build_abs_path, utcnow
+from superscore.validation import ValidationCode, ValidationResult
 
 logger = logging.getLogger(__name__)
 
@@ -717,7 +718,7 @@ class Client:
         sub_filled.creation_time = utcnow()
         return sub_filled
 
-    def validate(self, entry: Entry) -> bool:
+    def validate(self, entry: Entry) -> ValidationResult:
         """
         Validate ``entry`` is properly formed and able to be inserted into
         the backend.  Verifies the following in addition to the base model
@@ -727,13 +728,18 @@ class Client:
         """
         # Check for remaining placeholders
         if find_placeholders(entry):
-            return False
+            return ValidationResult(code=ValidationCode.UNFILLED_PLACEHOLDERS,
+                                    uuid=entry.uuid)
 
         # Verify that snapshots have valid collections in database
         if isinstance(entry, Snapshot):
             if not entry.origin_collection:
                 logger.debug("Snapshot does not have an origin collection")
-                return False
+                return ValidationResult(
+                    code=ValidationCode.ORIGIN_INVALID,
+                    uuid=entry.uuid,
+                    reason="Snapshot does not have an origin collection"
+                )
             if isinstance(entry.origin_collection, UUID):
                 linked_coll_uuid = entry.origin_collection
             else:
@@ -743,7 +749,11 @@ class Client:
             if not list(self.search(SearchTerm("uuid", "eq", linked_coll_uuid))):
                 logger.debug("Linked collection does not exist "
                              f"in database: {linked_coll_uuid}")
-                return False
+                return ValidationResult(
+                    code=ValidationCode.ORIGIN_INVALID,
+                    uuid=entry.uuid,
+                    reason="Snapshot's origin collection does not exist in database",
+                )
 
         # Use model validation
         return entry.validate()
--- a/superscore/client.py
+++ b/superscore/client.py
@@ -293,23 +293,28 @@ class Client:
         authorized = self.is_user_authorized(user)
 
         if not authorized:
+            logger.debug(f"User not authorized to edit entry: {entry.uuid}")
             return False
 
         # If backend does not allow writing, all else is moot
         if not self.backend.entry_writable(entry):
+            logger.debug(f"Backend does not allow entry editing: {entry.uuid}")
             return False
 
         # preventing editing of past entries adds additional constraints
         if not self.enable_editing_past:
             # Recent entries are allowed...
             if entry.uuid in self.recent_entry_cache:
+                logger.debug(f"Entry in recents cache, editing allowed: {entry.uuid}")
                 return True
 
             # while past entries are restricted
             if list(self.search(SearchTerm("uuid", "eq", entry.uuid))):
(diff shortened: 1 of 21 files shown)


tangkong:
I'm done adding extra features here, as always I ended up slightly outside my intended scope.  

I'll add tests and request reviews by eod today

superscore/client.py
```diff
@@ -328,6 +336,10 @@ def save(self, entry: Entry):
         else:
             entry_ids = [entry.uuid]
 
+        # Validate before saving
+        if not self.validate(entry):
```
ZLLentz, line 340: I was going to comment on this not being a boolean, but it does work as intended here because of how you extended the type. I'm a fan of this sort of implementation.

superscore/client.py
```diff
@@ -318,6 +324,8 @@ def save(self, entry: Entry):
         Should only save the parent information.  Children are to be references,
         and should not be modified by when their parent is saved.
         Perhaps this is to be left to the backend.
+        TODO:
+        - consider throwing error if this fails instead of being silent
```
ZLLentz, line 328: It is maybe a bit too silent- is the caller for the importing logic at least checking for success somehow?

I wonder if `save` should return the `ValidationResult`? Or if there's some other result it should be returning? Or maybe just good old-fashioned raise as you say here...

superscore/validation.py
```diff
@@ -0,0 +1,43 @@
...
+class ValidationCode(Enum):
+    """Terse Validation reasons"""
+    # TODO: consider flags for more thorough validation codes (combinations)
+    # Will only be relevant if validation routines don't terminate early
+    VALID = (auto(), "")
+    CYCLE_FOUND = (auto(), "Cycle in data hierarchy.  Parent can be re-reached from children")
+    ORIGIN_INVALID = (auto(), "Origin of Entry is either null or does not exist")
+    TYPE_ERROR = (auto(), "Entry has data of invalid type")
+    READBACK_INVALID = (auto(), "Entry has an invalid readback")
+    UNFILLED_PLACEHOLDERS = (auto(), "Entry has unfilled placeholders, please fill before saving")
```
ZLLentz, line 17: This is the first time I've seen tuples used like this in an enum- seems pretty snazzy. Are there any complications when it comes to creating enum members? Though I guess you wouldn't need the Enum(1) sort of syntax here.

superscore/widgets/window.py
```diff
@@ -205,6 +211,64 @@ def _window_context_menu(self, entry: Entry) -> QtWidgets.QMenu:
...
+            filename, _ = QtWidgets.QFileDialog.getOpenFileName(
+                parent=self,
+                caption='Select Entry Data',
+                filter='Json Files (*.json)',
+            )
+        if not filename:
+            return
+
+        try:
+            data = get_entry_from_path(filename)
```
ZLLentz, line 230: Are we validating and raising in here? If so, what's here is probably more than sufficient

ZLLentz (commented):
This looks good to me- maybe I give it a spin tomorrow and check some of the other parts of the app again? I'm not so interested in using this specific feature, but I am interested in the main parts of the app continuing to work.

superscore/client.py
tangkong: Not at all inspired by my desire to minimize the blast radius of the refactor 😆

superscore/client.py
tangkong: I agree that it's too silent, particularly in the spirit of allowing the GUI to be informative/responsive.  

In my mind the options for this (and for the validation results) are:
- fail silently
  - 🤫
- boolean representing success
  - simplest to implement, but least informative
  - requires intentional handling of result statuses (indented blocks)
- raise exception
  - requires indenting with try/except catch to avoid interrupting control flow
  - unexpected cases are handled "naturally", like all other exceptions
  - Best if we assume the data is good, and errors are unexpected
  - most jarring to the user
  - most pythonic?
- Message packet
  - allows passing more detailed information up the stack
  - requires special handling / boilerplate (unless we duck type as I did here)


But I also decided refactoring the save methods is scope creep here.

superscore/widgets/window.py
tangkong: The helper function will raise errors from the deserialization process, which includes apischema's validation.  We aren't running our own extra model validation however.  

I suppose I should be catching that error not a generic ValueError

superscore/validation.py
tangkong: I heavily prefer the `Enum.member` method of member creation, so I hadn't thought about that...
If you wanted the callable-esque syntax you'd need to pass the whole tuple

```python

In [6]: ValidationCode((1, ""))
Out[6]: <ValidationCode.VALID: 1>
```

ZLLentz: It's pretty rare you'd want to use that syntax so it's fine (especially here). I just wanted to understand all the tradeoffs.

ZLLentz (approved):
I'm going to approve this- if there's a lurking bug or usability issue we'll catch it one of the deployed builds.

tangkong:
The old "test it in prod" approach I see.

PR #153, opened 2026-04-16
ENH: add batch restore capability

tangkong:
## Description
The last feature for real

- Allows user to restore a snapshot in chunks, rather than all at once through the `RestoreDialog`
  - Refactors the table to use a proper model with button delegates for synchronizing the data model and list of entries
 
TODO:
- [ ] Consider adding this ability to the client, but in a restore batch - wait loop

## Motivation and Context
Apparently we're might restore enough PVs to crash an IOC, and we'd like to not do that

## How Has This Been Tested?
Added some tests, did some mild interactive testing

This PR

image

## Pre-merge checklist

### DIFF
--- a/superscore/widgets/page/restore.py
+++ b/superscore/widgets/page/restore.py
@@ -69,7 +69,23 @@ class RestoreDialogModel(QtCore.QAbstractTableModel):
     def columnCount(self, parent: Optional[QtCore.QModelIndex] = None):
         return 3
 
-    def data(self, index: QtCore.QModelIndex, role: int) -> Any:
+    def headerData(
+        self,
+        section: int,
+        orientation: QtCore.Qt.Orientation,
+        role: int = QtCore.Qt.DisplayRole
+    ) -> Any:
+        """
+        Returns the header data for the model.
+        Currently only displays horizontal header data
+        """
+        if role != QtCore.Qt.DisplayRole:
+            return
+
+        if orientation == QtCore.Qt.Horizontal:
+            return self.headers[section]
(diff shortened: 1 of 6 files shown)


superscore/widgets/page/restore.py
```diff
@@ -54,50 +56,127 @@ def toggle_live(self):
...
+    def flags(self, index: QtCore.QModelIndex) -> QtCore.Qt.ItemFlags:
+        """
+        Returns the item flags for the given ``index``.  Set delegates editable
+        """
+        if index.column() != 2:
+            return QtCore.Qt.ItemIsEnabled
+
+        return QtCore.Qt.ItemIsEnabled | QtCore.Qt.ItemIsEditable
+
+    def removeRow(self, row: int, parent: QtCore.QModelIndex = ...) -> bool:
```
ZLLentz: A lot of the return value type hints in this class are imprecise, but I think this is the only one that's wrong

ZLLentz (dismissed):
This seems like a reasonable implementation to me. It handles the requested scope well and doesn't assume extra scope.

I'm pretty sure my only quibbles are:

- imprecise and incorrect type annotations
- no docstrings

tangkong:
Good call out.  I'm drawing the laziness line at methods that just re-implement the qt interfaces, but I added some docstrings.

ZLLentz:
Yeah I'm not too worried about those- the one that stuck out to me was the one where it didn't match

ZLLentz (approved):

